Issue #257, opened 2014-03-16
RemovedInDjango18Warning deprecation warnings


blueyed:
I see two RemovedInDjango18Warning warnings (from Django 1.7alpha) with autocomplete-light:

 - [ ] …/autocomplete-light/autocomplete_light/forms.py:266: RemovedInDjango18Warning: Creating a ModelForm without either the 'fields' attribute or the 'exclude' attribute is deprecated - form ModelForm needs updating
      attrs)
 - [ ] …/autocomplete-light/autocomplete_light/forms.py:449: RemovedInDjango18Warning: Calling modelform_factory without defining 'fields' or 'exclude' explicitly is deprecated
      return django_modelform_factory(model, **kwargs)

linaskilius:
I'd love to see this patch released as well.  Hopefully I count as some added user demand :)

jpic:
Fix released in 2.0.8.

It's going to be way cool when Django 1.8 LTS is released we'll be able to start removing legacy code.

PR #286, opened 2014-06-06
Add empty `exclude` to default Meta class in modelform_factory

blueyed:
This is deprecated since Django 1.7 and fails in master.

Fixes https://github.com/yourlabs/django-autocomplete-light/issues/257.


andybak:
One final observation. As well as updating the FAQ - the code examples should probably include fields or exclude. i.e. http://django-autocomplete-light.readthedocs.org/en/stable-2.x.x/tutorial.html#autocomplete-light-modelform-factory-shortcut-to-generate-modelforms-in-the-admin

jpic:
Agreed, excellent idea.

Issue #247, opened 2014-01-31
Please add Django 1.7 compatibility


lorddaedra:
```python3
(bemyco3) iMac-Alexander:bemyco lorddaedra$ ./manage.py
Traceback (most recent call last):
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/autocomplete_light/registry.py", line 205, in __getitem__
    return super(AutocompleteRegistry, self).__getitem__(name)
KeyError: 'CityAutocomplete'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/django/core/management/__init__.py", line 427, in execute_from_command_line
    utility.execute()
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/django/core/management/__init__.py", line 391, in execute
    django.setup()
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/django/__init__.py", line 21, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/django/apps/registry.py", line 111, in populate
    app_config.ready()
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/django/contrib/admin/apps.py", line 22, in ready
    self.module.autodiscover()
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/django/contrib/admin/__init__.py", line 23, in autodiscover
    autodiscover_modules('admin', register_to=site)
  File "/Users/lorddaedra/Envs/bemyco3/lib/python3.3/site-packages/django/utils/module_loading.py", line 58, in autodiscover_modules
    import_module('%s.%s' % (app_config.name, module_to_search))
  File "/usr/local/Cellar/python3/3.3.3/Frameworks/Python.framework/Versions/3.3/lib/python3.3/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1584, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1565, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1532, in _find_and_load_unlocked
[...]

PR #293, opened 2014-06-20
Use django-1.7 apps registry.

g1itch:
To deal with django-1.7 `AppConf`s in `settings.INSTALLED_APPS`. Maybe it's a part of the #247.

### DIFF
--- a/autocomplete_light/registry.py
+++ b/autocomplete_light/registry.py
@@ -180,13 +180,15 @@ def _autodiscover(registry):
     import copy
     from django import get_version
     from django.conf import settings
-    from django.apps import apps
     from django.utils.importlib import import_module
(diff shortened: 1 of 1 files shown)


blueyed:
Just for reference: this was never added to v2, but seems to be addressed there differently via 5b1ed105 (Fix #347 Django 1.7 AppConfig support).

jpic:
Correct !

blueyed:
@jpic 
Please elaborate.. what is correct?
I have cherry-picked this in a local branch already and could provide a PR - but I am not sure what you mean.

blueyed:
I see. You've replied to my comment before I've added the 2nd paragraph.
All fine then.

Issue #294, opened 2014-06-20
Ampersand character (&) cannot be typed in the autocomplete input


yliharma:
If you try to type the character _&_ (which in my qwerty european keyboard is SHIFT+6) nothing happens.
I probably found the solution to that by adding a condition in autocomplete.js, row 453 in yourlabs.Autocomplete.prototype.move, that checks if the key pressed is the shift key (apparently the browser mistakes Shift for keyUp...)

Your code:
```javascript
    // If not KEY_UP or KEY_DOWN, then return.
    if (e.keyCode == 38) var way = 'up';
    else if (e.keyCode == 40) var way = 'down';
    else return;
```
becomes:

```javascript
    // If not KEY_UP or KEY_DOWN, then return.
    if (e.keyCode == 38 && !e.shiftKey) var way = 'up';
    else if (e.keyCode == 40) var way = 'down';
    else return;
```

aidanlister:
Is this fixed now? I can't find a commit referencing it

yliharma:
Your pull request has been merged!

blueyed:
@aidanlister 
It has been committed in 51e20fa31189d47a05625335dbad96480bc82def.

blueyed:
@aidanlister @yliharma 
Can you please test if the following patch/PR still/also fixes it for you:
https://github.com/yourlabs/django-autocomplete-light/pull/398

It should be cleaner that way, but I want to not cause a regression here.

PR #306, opened 2014-07-14
Fixes a bug where ampersand cannot be typed on Chrome/Mac

aidanlister:
Fixes issue #294, fix by @yliharma

### DIFF
--- a/autocomplete_light/static/autocomplete_light/autocomplete.js
+++ b/autocomplete_light/static/autocomplete_light/autocomplete.js
@@ -450,7 +450,7 @@ yourlabs.Autocomplete.prototype.move = function(e) {
     }
 
     // If not KEY_UP or KEY_DOWN, then return.
-    if (e.keyCode == 38) var way = 'up';
+    if (e.keyCode == 38 && !e.shiftKey) var way = 'up';
     else if (e.keyCode == 40) var way = 'down';
     else return;
 


blueyed:
Please test if https://github.com/yourlabs/django-autocomplete-light/pull/398 also fixes this.

yliharma:
Nope, #398 broke the ampersand again :(

jpic:
Is it fixed in 2.1.0rc2 ?

Thanks for your feedback B)

yliharma:
It works...but I'm not sure which version I'm using.
I upgraded the package running `pip install -U -e git+https://github.com/yourlabs/django-autocomplete-light.git#egg=autocomplete_light` in my virtualenv and in the output there was `django-autocomplete-light 2.1.0rc2 is already the active version in easy-install.pth` but if I open a python shell and import the module, the output of `autocomplete_light.VERSION` is `(1, 6, 10, 'final', 0)`...

jpic:
Lol nice one, this is because of django 1.9 support xD

https://github.com/yourlabs/django-autocomplete-light/blob/76a617b4aaffd36ae61c29748fe480cccdc9cffb/autocomplete_light/__init__.py

What you're seeing here is Django's version !

Gotta fix this

aidanlister:
pip freeze to see your version info

yliharma:
Aidan it's not easy to understand...the output is  `-e git+https://github.com/yourlabs/django-autocomplete-light.git@76a617b4aaffd36ae61c29748fe480cccdc9cffb#egg=django_autocomplete_light-master` where probably the portion between @ and # is the changeset number...but then I would have to look for this commit: boring!!! I can tell you that it's not the last commit.

jpic:
Ok cd in yourenv/src/django_autocomplete_light and do git pull origin master.

PEP 396 is implemented now so autocomplete_light.__version__ should be available.

yliharma:
Perfect! Thank you :+1:

PR #334, opened 2014-11-14
Little changes to avoid bad style to underlying html elements

madEng84:
Little modifications to remove strange effect to underlying html elements in django admin

You can see what happens before 
schermata 2014-11-14 alle 12 53 38

and after my modifications
schermata 2014-11-14 alle 12 54 10

### DIFF
--- a/autocomplete_light/static/autocomplete_light/autocomplete.js
+++ b/autocomplete_light/static/autocomplete_light/autocomplete.js
@@ -506,7 +506,7 @@ yourlabs.Autocomplete.prototype.fixPosition = function() {
 
     this.input.parents().filter(function() {
         return $(this).css('overflow') === 'hidden';
-    }).first().css('overflow', 'visible');
+    }).first().css('overflow', 'visible').addClass('autocomplete-light-clearfix');
 	
     this.box.insertAfter(this.input).css(
             {top: pos.top + pos.height, left: pos.left});
--- a/autocomplete_light/static/autocomplete_light/style.css
+++ b/autocomplete_light/static/autocomplete_light/style.css
@@ -153,6 +153,19 @@ input.autocomplete.xhr-pending {
     display: none;
 }
 
+/* These lines are important to avoid style change to the underlying divs */
+.autocomplete-light-clearfix:before,
+.autocomplete-light-clearfix:after {
+    content: " "; 
+    display: table; 
+}
+.autocomplete-light-clearfix:after {
+    clear: both;
+}
+.autocomplete-light-clearfix {
+    *zoom: 1;
+}
+
 /* grappelli specific */
 fieldset.grp-module .grp-row {
     overflow: visible !important;


jualvarez:
This *fix* breaks the display of inline forms in django 1.7. It shifts the whole row by one `<td>`. Removing the `content: " "` from the style seems to fix the problem. Not sure about other consequences.

image

sbaum:
Some weeks ago I've stumbled across the same problem that jualvarez has reported yesterday. Unfortunately I've forgotton to report it here. :-/
On my django project I've applied the same fix that jualvarez has mentioned: I've removed the line `content: " "` from style.css. But I can't tell if this has any negative side effects.
So I can confirm that the merged pull request #334 has broken the display of inline forms in Django 1.7.

jpic:
Could Younes submit a PR with your fix ?

sbaum:
Perhaps madEng84 should tell us first if there are any negative side effects if the `content: " "` line gets removed. I think he can tell, as he is the author of the regarding pull request.

madEng84:
Hi, 
i have added the clearfix not to broke the underlying divs
I took that code from here (the second to last):
http://css-tricks.com/snippets/css/clear-fix/

Reading better that link, i see i was wrong because i had to insert 'content : ""' and not 'content:" "'
Can you try if deleting that space in content field it seems the same in Django 1.7?
Excuse me but i'm very busy in this period.

Thanks a lot
Regards

jualvarez:
@madEng84 `content: ""` has the same effect. I haven't investigated further on why this happens. I'll update if I can find another workaround. Removing the `content: " "` style altogether breaks the output for your case too? Thanks!

madEng84:
@jualvarez 
Removing the content: " " style, it breaks the output.
If i replace it with content: "" it works good instead.
schermata 2015-02-23 alle 18 22 49
schermata 2015-02-23 alle 18 22 37
schermata 2015-02-23 alle 18 30 55

jpic:
What browsers and django admin themes are you talking about ?

sbaum:
The problem reported by jualvarez and me is affecting at least Firefox and Chrome browsers (latest versions) running on Ubuntu Linux and Windows 7, as well as Internet Explorer 11 running on Windows 7. I'm using stock django admin, no custom theme.

@madEng84 Couldn't we replace

    .autocomplete-light-clearfix:before,
    .autocomplete-light-clearfix:after {
        content: " ";
        display: table; 
    }
    .autocomplete-light-clearfix:after {
        clear: both;
    }
    .autocomplete-light-clearfix {
        *zoom: 1;
    }

with

    .autocomplete-light-clearfix:after {
        content: "";
        display: table; 
        clear: both;
    }

? That's the latest version of the clearfix, as it is stated on the bottom of the page you've mentioned (http://css-tricks.com/snippets/css/clear-fix/, search for "Update August 2012"). That shorter CSS does not cause the problems we are experiencing. If I'm understanding it right, we would only lose IE 6/7 compatibility.

jpic:
We can let IE 6 and 7 die as it deserve.

madEng84:
Excuse me @jpic 
I'm using Django<1.7 and Chrome 40.0.2214.115 (64-bit)

sbaum:
I've just submitted the corresponding PR #374 .

PR #341, opened 2014-11-29
Fixed  #340 Added compatibility with django-hstore

### LINKED ISSUE #340: Add compatibility with django-hstore
This project is currently incompatible with django-hstore schema mode. hstore introduces it's own [VirtualField](https://github.com/djangonauts/django-hstore/blob/master/django_hstore/virtual.py). I've demonstrated the issue [here](https://github.com/bufke/autocomplete_hstore). 

autocomplete expects ._meta.virtual_fields to contain a fk_field property [here](https://github.com/yourlabs/django-autocomplete-light/blob/stable/2.x.x/autocomplete_light/forms.py#L358). However the hstore VirtualField does not have a fk_field.

```
for field in meta.model._meta.virtual_fields:
    ...
    required=not meta.model._meta.get_field_by_name(field.fk_field)
```

A check `if hasattr(field, 'fk_field'):` fixes the issue and I can't see it causing any harm. Would you consider a pull request to support this? If you feel this is better solved on django-hstore's end please let me know. I'd love to get a fix accepted on one of the projects.

See [issue](https://github.com/djangonauts/django-hstore/issues/86) on django-hstore too.

bufke:


jpic:
Are you still working on this ?

We'd like to rename branches (#349) and that will close existing pull requests, so you'd have to create a new one.

For tests, check how tests for django-generic-m2m look like: 
https://github.com/yourlabs/django-autocomplete-light/blob/stable/2.x.x/autocomplete_light/tests/generic_m2m.py

bufke:
Hey - Yeah I'd like to get this done with a unit test. That link is broken - did the file get renamed?

blueyed:
The test case appears to have been removed in 732488f lately.

jpic:
Ok so either we can stack a case in autocomplete_light/example_apps/basic and test it in autocomplete_light/forms like other high level native features either add a test in autocomplete_light/tests/hstore.py if you don't want to test the whole processes.

jpic:
> The test case appears to have been removed in 732488f lately.

That file was left over from 1.x ... removing it should have been done for 2.0 stable release already.

jpic:
I added a test in the fix/hstore branch but it doesn't fail.

It would be great if you explained your use-case, what fails and how, or contributed to that branch so that:

- the test fails without your PR
- the test passes with your PR

Thanks

bufke:
Thank you for your attention to this. However I have stopped using django-hstore and cannot commit time to creating a suitable unit test. Given that django-hstore does not support 1.8 and that 1.8 has native (if less featured) postgres hstore support it might be fine to just close this.

jpic:
I got a test in: https://github.com/yourlabs/django-autocomplete-light/commit/72642390b476a7fd195b8b90d5bcf650fab5361d

But it doesn't fail

jpic:
Couldn't reproduce, even in test case, closing.

Please reopen if it reoccurs.

PR #345, opened 2014-12-15
[v1] Fix for positioning of MultipleChoiceWidget drop-down

eduvik:
This fixes an issue I had with the 1.4 branch where each subsequent drop down appears lower down the screen than it should, resulting in it being unusable after a few items have been entered. Not tested with 2.x

### DIFF
--- a/autocomplete_light/static/autocomplete_light/autocomplete.js
+++ b/autocomplete_light/static/autocomplete_light/autocomplete.js
@@ -509,7 +509,7 @@ yourlabs.Autocomplete.prototype.fixPosition = function() {
     }).first().css('overflow', 'visible').addClass('autocomplete-light-clearfix');
 	
     this.box.insertAfter(this.input).css(
-            {top: pos.top + pos.height, left: pos.left});
+            {top: pos.height, left: pos.left});
 }
 
 // Proxy fetch(), with some sanity checks.


jpic:
Awesome ! You now have contributor credentials on GitHub which allows you to merge your change when you're ready to go ahead and release. You'll need to send me your PyPi username too to be able to upload the release. Don't forget to add yourself to AUTHORS ;)

blueyed:
@jpic 
While I am happy that you are extending the team of contributors, I suggest that releases get done / reviewed by you only.
Also, I do not think every small / non-critical fix should make it into a release immediately.

@eduvik 
> Not tested with 2.x

Did you mean to base this PR on version 2 then?

Apart from that, it would be great to have a test for this fix / PR.

jpic:
> Apart from that, it would be great to have a test for this fix / PR.

Agreed, @eduvik could you provide an example ?

askvictor:
I had a go at setting up the test project and creating a test... but I couldn't get the 1.4 test project running, and couldn't replicate in 2.0 - my current code which this PR fixes uses the old method of specifying the widget, so I couldn't create a setup similar to mine the time I had. I don't have any more time to spend on this at the moment; when upgrade to 2.0 in my project, I'll see if the issue is still there,  and once I've moved to the new style, I'll see if I can create  a test for this then. In the meantime, I'm just going to use my fork of 1.4 for my site.

jpic:
I couldn't get it to work.

blueyed:
@jpic 
It does not appear to affect v2, but v1 only, according to what @askvictor said:

> couldn't replicate in 2.0

So we should probably cherry-pick this into the stable/1.x.x branch, given that both @askvictor and @eduvik have this problem in 1.x - also when nobody comes up with a test case.

Does this make sense?

jpic:
Merge on stable/1.x.x sounds reasonable - if we find one 1.x.x user who's not using a fork (forking for this sounds insane though, you could have just overridden that method or even the entire script ...).

blueyed:
Ok. Pushed as 06bf9cc to the stable/1.x.x branch.

jpic:
Thanks @blueyed

PR #350, opened 2015-01-06
Remove old_test_project

blueyed:
Is this OK to merge?
The old_test_project does not appear to get used anymore.

It seems to have been created in 82e8a9f, and should maybe never have been committed / maintained later at all.


jpic:
Agreed ! Let's move on.

RIP old_test_project, and test_remote_project can probably die too - I
never tried it with v2 :)

PR #351, opened 2015-01-06
Improve/fix test.sh

blueyed:
- Update Django versions, use less-than and more-than comparisons with
   `pip install`.
 - Do not install Django dep with test project requirements, to avoid an
   error from pip with duplicated requirements/options given.
 - Do not install specific django-cities-light version, handled by
   test_project/test_requirements.txt.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -9,8 +9,8 @@ WORKSPACE="${WORKSPACE:-$(pwd)}"
 CLEAN_VIRTUALENV="${CLEAN_VIRTUALENV:-0}"
 DJANGO_TAGGIT="${DJANGO_TAGGIT:-1}"
 DJANGO_GENERIC_M2M="${DJANGO_GENERIC_M2M:-1}"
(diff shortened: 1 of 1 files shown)


test.sh
```diff
@@ -18,9 +18,13 @@ XVFB_FLAGS="${XVFB_FLAGS:-}"
 ENV_PATH="$WORKSPACE/test_env"
 
 # Get real django version
-[ "$DJANGO_VERSION" = "1.4" ] && DJANGO_VERSION="1.4.10"
-[ "$DJANGO_VERSION" = "1.5" ] && DJANGO_VERSION="1.5.5"
-[ "$DJANGO_VERSION" = "1.6" ] && DJANGO_VERSION="1.6.1"
+case "$DJANGO_VERSION" in
+    1.4) django_dep='django>1.4,<1.5' ;;
```
jpic: I think <1.5 alone would work here, it works in .travis.yml

PR #357, opened 2015-01-20
Remove tests/generic.py, not included anymore

blueyed:
It was for autocomplete/generic.py which is now tested in forms.py
like all basic forms.

This is what was probably meant to be removed in 2a8a449.


blueyed:
Cherry-picked into master in 051bcb7 - the commit message was not really correct.

PR #358, opened 2015-01-20
remove non-ascii characters from text_widget.js

askholme:
This removes a non-ascii whitespace like character from text_widget.js (the character is similiar to alt+space on OS X). Though not noticed on a normal glance those characters does not play nice with minification tools


blueyed:
Thanks! Cherry-picked into master at 42757f6.

I've also removed the trailing whitespace while at it.

PR #359, opened 2015-01-20
Minor code cleanup

blueyed:


jpic:
Apparently this only breaks configuration where at least one of django-generic-m2m or django-taggit is not installed.

blueyed:
Addressed in https://github.com/blueyed/django-autocomplete-light/commit/c8e29f78911a7dac303557cf4fbdc077121ea249

PR #305, opened 2014-07-09
WIP: Use choices from model admin field queryset

blueyed:
Ref: https://github.com/yourlabs/django-autocomplete-light/issues/300

Followed up in #360.

PR #360, opened 2015-01-20
Model choices from field queryset

blueyed:
Followup of #305 against master.

### DIFF
--- a/autocomplete_light/fields.py
+++ b/autocomplete_light/fields.py
@@ -30,11 +30,11 @@ class FieldBase(object):
                     autocomplete_js_attributes, extra_context)
         kwargs['widget'] = widget
 
-        parents = super(FieldBase, self).__self_class__.__bases__
-        if ((forms.ModelChoiceField in parents or
-                forms.ModelMultipleChoiceField in parents)
-                and isinstance(self.autocomplete.choices, QuerySet)):
-            kwargs['queryset'] = self.autocomplete.choices
+        if 'queryset' not in kwargs:
+            parents = super(FieldBase, self).__self_class__.__bases__
+            if (forms.ModelChoiceField in parents or
+                    forms.ModelMultipleChoiceField in parents):
+                kwargs['queryset'] = self.autocomplete.choices
 
         super(FieldBase, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


autocomplete_light/fields.py
```diff
@@ -30,11 +27,11 @@ def __init__(self, autocomplete=None, registry=None, widget=None,
...
-        parents = super(FieldBase, self).__self_class__.__bases__
-        if ((forms.ModelChoiceField in parents or
-                forms.ModelMultipleChoiceField in parents)
-                and isinstance(self.autocomplete.choices, QuerySet)):
-            kwargs['queryset'] = self.autocomplete.choices
+        if 'queryset' not in kwargs:
+            parents = super(FieldBase, self).__self_class__.__bases__
+            if (forms.ModelChoiceField in parents or
+                    forms.ModelMultipleChoiceField in parents):
+                kwargs['queryset'] = self.autocomplete.choices
```
jpic: Well I really dislike how I coupled Model stuff into FieldBase, can't find out what was the reason for that ... so should we consider moving that part into ModelChoiceField ?

autocomplete_light/fields.py
```diff
@@ -88,13 +85,24 @@ class MultipleChoiceField(ChoiceField, forms.MultipleChoiceField):
...
+    def __init__(self, autocomplete=None, registry=None, widget=None,
+            widget_js_attributes=None, autocomplete_js_attributes=None,
+            extra_context=None, *args, **kwargs):
+
+        super(ModelChoiceField, self).__init__(autocomplete, registry, widget,
+                widget_js_attributes, autocomplete_js_attributes,
+                extra_context, *args, **kwargs)
+
+        # Store ourself in the autocomplete, for AutocompleteModel.choices.
+        self.autocomplete.model_field = self
```
jpic: While I understand how convenient it is to have the model field instance in the autocomplete, it seems like we're coupling Autocompletes to Fields by doing that. I'm not firmly against it, but in that case shouldn't we find a more global way, that would support say django.forms.ModelChoiceField.queryset ?

Meanwhile, should we consider setting self.autocomplete.choices here directly ?

jpic: Oh and be careful, it should be self.autocomplete.form_field here i guess, i'd assume self.autocomplete.model_field to be an instance of the fk/m2m model field.

jpic:
After another proof reading I'm really suprised it works. But that did totally open my eyes, we can hack a registered autocomplete from form fields after all, it doesn't look very safe, what happens if an autocomplete is used in two different form fields ?

> self.autocomplete.model_field = self

What would be pretty neat for v3 though, would be to get completely rid of Autocomplete classes and rely on form fields instead. But that means finding a good way to get the form field back in the AutocompleteView ...

Issue #239, opened 2014-01-20
Use ModelAdmin.search_field as default for ModelComplete.search_fields


blueyed:
While the following appears to work in the registry, I think it would be a nice default, instead of using DEFAULT_SEARCH_FIELDS (which may not be available):

    autocomplete_light.register(Model, search_fields=ModelAdmin.search_fields)

I don't know if that's easily possible, or another attempt has to be used in the registry to make it work - I just wanted to leave it here as an idea.

jpic:
Closing in favor of PR #361

PR #301, opened 2014-07-08
register: default to admin `search_fields`, fallback to 'name'/'title'

blueyed:
Fixes https://github.com/yourlabs/django-autocomplete-light/issues/239

All tests still pass, but new ones should get added probably?!

PR #361, opened 2015-01-20
register: default to admin `search_fields`, fallback to 'name'/'title'

blueyed:
Followup of https://github.com/yourlabs/django-autocomplete-light/pull/301 against master.

Fixes https://github.com/yourlabs/django-autocomplete-light/issues/239.

### DIFF
--- a/autocomplete_light/registry.py
+++ b/autocomplete_light/registry.py
@@ -21,12 +21,14 @@ It looks like this:
 """
 import six
 
+from django.contrib import admin
 from django.db import models
 
-from .autocomplete import AutocompleteModelBase, AutocompleteInterface
-from .exceptions import (AutocompleteNotRegistered,
-                         AutocompleteArgNotUnderstood,
-                         NoGenericAutocompleteRegistered)
+from .autocomplete import AutocompleteInterface, AutocompleteModelBase
+from .exceptions import (AutocompleteArgNotUnderstood,
+                         AutocompleteNotRegistered,
+                         NoGenericAutocompleteRegistered,
+                         UnconfiguredSearchFieldsException)
 
 try:
     from django.utils.module_loading import autodiscover_modules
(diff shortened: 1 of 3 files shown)


blueyed:
This should be called after all other apps have registered any admin classes for their models, so that `search_fields` is actually available.

That could be achieved by having it after django.contrib.admin in INSTALLED_APPS, but the documentation says that it's required to be in front.

Could search_field become a property, that looks up the default via the admin's registry on demand/later?

jpic:
> This should be called after all other apps have registered any admin classes for their models, so that search_fields is actually available.

Good point ^^

> That could be achieved by having it after django.contrib.admin in INSTALLED_APPS, but the documentation says that it's required to be in front.

Well the admin needs forms, and forms need autocomplete_light_registries to be loaded for any autocomplete action to happen.

> Could search_field become a property, that looks up the default via the admin's registry on demand/later?

This could work, after all, search_fields powers choices_for_request() which is used by autocomplete_html(), which is called by AutocompleteView. In theory, there is no need for search_fields to be configured upfront. 

You'd like to try making AutocompleteModel.search_fields an `@property` method ? Makes sense to me.

blueyed:
> You'd like to try making AutocompleteModel.search_fields an @property method ? Makes sense to me.

Yes.

The drawback I see is that it would not result in errors (`UnconfiguredSearchFieldsException`) during startup, but during runtime.

Is there another possibility, e.g. a hook / signal after all apps have been loaded?

jpic:
> Is there another possibility, e.g. a hook / signal after all apps have been loaded?

Actually, there is a new feature which I haven't yet looked at in details: https://docs.djangoproject.com/en/1.7/ref/checks/

Would that work ?

blueyed:
That would work probably.

jpic:
Sounds like the next release is all yours!
Rock'on

autocomplete_light/registry.py
```diff
@@ -173,12 +176,20 @@ def _register_model_autocomplete(self, model, autocomplete=None,
 
         if base.search_fields is None and 'search_fields' not in kwargs:
             try:
-                model._meta.get_field('name')
-            except:
-                raise Exception('Add search_fields kwargs to .register(%s)'
-                                % model.__name__)
-            else:
-                kwargs['search_fields'] = ['name']
+                search_fields = admin.site._registry[model].search_fields
```
jpic: We might be using the wrong site though, we need the admin_site from there: https://github.com/django/django/blob/stable/1.8.x/django/contrib/admin/widgets.py#L241

PR #362, opened 2015-01-21
Try PhantomJS on travis (again)

blueyed:
Cherry-picked from 2160ae3.

Conflicts:
	.travis.yml
	autocomplete_light/tests/widget.py


jpic:
Maybe it would be good to load the driver according to some environment variable, so that you could choose which driver you want without changing the code.

blueyed:
Ack.
When I get at integrating / using py.test again, this would be done / provided via `--splinter-webdriver`.

Here, we could use the process environment and maybe have different entries in the scripts.
Also, different defaults could be used for travis / tox / test.sh.

blueyed:
This should get rebased on top of #366 and use the cache for the phantomjs download.

PR #367, opened 2015-01-21
Use Django's TestCase

blueyed:
This should result in less broken tests (https://github.com/pytest-dev/pytest-django/pull/197#issuecomment-70806617), and will make it easier to use pytest.


blueyed:
@jpic 
Ok to merge?

PR #387, opened 2015-04-02
Improve speed with `login`: use cookie from self.client

blueyed:
This is from an older branch I had locally.


blueyed:
This has been added in eadcd23.

PR #392, opened 2015-04-03
Fix RemovedInDjango19Warning on flatatt

SaeX:


blueyed:
Thanks, committed to master in a9a8cc3.

PR #396, opened 2015-04-08
To prevent RemovedInDjango18Warning on modelform_factory of forms

ebertti:
To prevent this warning message: RemovedInDjango18Warning: Calling modelform_factory without defining 'fields' or 'exclude' explicitly is deprecated


jpic:
No, we're not merging this security vulnerability.

If Django requires 'fields' or 'exclude' to be defined explicitely, it's for security reasons.

Issue #394, opened 2015-04-07
left parenthesis ("(") and ampersand ("&") characters are ignored


jindrichjindrich:
Due to the fact that event.keyCode for "(" is the same as for a KEY_DOWN (40) and event.keyCode for "&" is the same as for a KEY_UP (38) those characters are ignored when an attempt to enter them in a search string is made.

I solved it by replacing each occurrence of e.keyCode with _keyCode(e) in autocomplete.js.

_keyCode is a function:

function _keyCode(e) { return (e.charCode + 1) * e.keyCode; }

blueyed:
That might be a more generic fix for https://github.com/yourlabs/django-autocomplete-light/issues/294 (which is fixed in master).

blueyed:
Can you please try the fix from https://github.com/yourlabs/django-autocomplete-light/pull/398 ?

I think that this fix goes into the same direction, by using `e.which` instead of `e.keyCode` (which prefers `e.charCode`, too).

yliharma:
Nope, #398 broke the ampersand again!

blueyed:
@yliharma 
Thanks for reporting back.  I see now what the issue is, and it's not related to Mac, but Webkit in general (I can confirm it here).

@jindrichjindrich 
The issue is fixed in Git / the current rc already.

I will add a comment to the code, and use the `charCode == 0` check instead, which appears to be more sensible to me.

yliharma:
Have you tried it with non-webkit browsers, like Firefox or our "old friend" Internet Explorer?

blueyed:
It works with Firefox, but haven't tried IE.

yliharma:
Perfect for me :+1:

jpic:
Excellent, fix released in 2.1.0rc3.

PR #398, opened 2015-04-08
Use event.which (managed by jQuery) instead of e.keyCode

blueyed:
Fixes https://github.com/yourlabs/django-autocomplete-light/issues/394

Needs to get confirmed / more testing.


blueyed:
Closing this.  Using `which` is not the right approach here.

For some reason keyCode gets reported as 38 for ampersands, although that should be charCode then (only).  And using `.which` makes it only worse then.

PR #400, opened 2015-04-10
WIP: system checks (initially for incompatible field and widget)

blueyed:


jpic:
I think we're going to end up with a lot of tests, so maybe we should use pytest-django's feature instead of hardcoding in test apps.

Maybe try adding a couple of other cases and see how refactoring looks like ?

Case ideas:

- overriding widget in formfield_override should not raise a warning if also overriding form field class,
- overriding with a widget that is a child autocomplete_light.FieldBase should not raise a warning

jpic:
Do you want to rebase and merge that one ?

Issue #379, opened 2015-02-26
Difficult to subclass FieldBase for custom ModelChoiceFields


smcoll:
i'm trying to build a custom autocomplete multiple model choice field.  i started like this:

```
class SortedModelMultipleChoiceField(autocomplete_light.FieldBase, SortedMultipleChoiceField):
    pass
```

But that results in a `TypeError: __init__() takes at least 2 arguments (2 given)` because 'queryset' is not in `kwargs`.

This is because `parents` in  https://github.com/yourlabs/django-autocomplete-light/blob/master/autocomplete_light/fields.py#L33 doesn't return `forms.ModelMultipleChoiceField` in my case; it only returns `(<class 'autocomplete_light.fields.FieldBase'>, <class 'sortedm2m.forms.SortedMultipleChoiceField'>)` even though `sortedm2m.forms.SortedMultipleChoiceField` is a subclass of `forms.ModelMultipleChoiceField`

Using `__mro__` instead of `__bases__` would solve the check.

For the time being, i'm just declaring the field this way:

```
class SortedModelMultipleChoiceField(autocomplete_light.FieldBase, SortedMultipleChoiceField, forms.ModelMultipleChoiceField):
    """
    forms.ModelMultipleChoiceField is required as a parent class, even though 
    it's a parent class of SortedMultipleChoiceField
    """
    pass
```

jpic:
If that does the trick it looks ok but we need a regression test !

jpic:
Could you make PR and test to ensure this gets in next release please ? (~2 weeks)

jpic:
Fix merged and released in 2.1.0rc3.

Please confirm 2.1.0rc3 fixes it.

Thanks a lot for your contribution !

smcoll:
For the record, with this fix in addition to #331 , making an autocomplete version of sortedm2m's SortedMultipleChoiceField was super clean:

```
import autocomplete_light
from sortedm2m.forms import SortedMultipleChoiceField


class SortedModelMultipleChoiceField(autocomplete_light.FieldBase, SortedMultipleChoiceField):
    """
    Autocomplete ModelMultipleChoiceField with SortedM2M support
    """
    widget = autocomplete_light.MultipleChoiceWidget
```

carlos22:
Since v3 is completely different, is there a similar approach to use dal with sortedm2m?

PR #401, opened 2015-04-11
fix and regression tests for #379

smcoll:
Potential fix for #379.  Maybe the regression tests are overkill...


jpic:
Sounds good for me. Thanks for your contribution !

PR #405, opened 2015-04-15
Django 1.8 is also supported

SaeX:


jpic:
Thanks !

Even Django 1.9 is supported ;)

PR #407, opened 2015-04-16
AutocompleteModel.choices_for_values() retains ordering of values

smcoll:
To start with, added a failing test which asserts the AutocompleteModel.choices_for_values() does not modify the ordering of the values.

Otherwise, no implementation yet; this PR is not complete.


jpic:
Pretty good, go ahead and get the test to pass so that we can have this in 2.1.0 release B)

Thanks !

PR #409, opened 2015-05-01
Prefer mousedown and specify input

SebCorbin:
Hello,
Thanks to your amazing package I've been able to fulfill my most advanced UX desires, you made some nice work here :)

Although I've found two small things that bothered me:

1. Prefering mousedown instead of click event allows to work on data already processed by autocomplete
2. I've used readonly inputs in my choice template, but they get picked up at initialisation, so I needed to specify the original input


jpic:
But click is touch-compatible wheras mousedown is not am I correct ?

jpic:
Touch suport is handled by #418 , merging.

PR #412, opened 2015-05-05
Get the primary key name from choices so does not have to be id

mhuailin:
Right now if you attempt to use autocomplete-light on a table where the primary key is not id, you get a id field error.


jpic:
Closing in favor of #417, which contains your PR + some tests.

jpic:
Could you try with 2.2.0rc1 which was [released yesterday](http://blog.yourlabs.org/post/119860038918/django-autocomplete-light-2-2-0rc1-release-notes) ? 

Thanks.

PR #414, opened 2015-05-14
travis: write PIP_WHEELHOUSE into travis_pip_install script

blueyed:
tox 2.0 won't use the parent env anymore by default, and we can just
write the env var into the script.


blueyed:
Merged in a7053db.

PR #417, opened 2015-05-17
Fix #412: Support models with a pk different than "id" and non-numeric.

jpic:
This is #412 with tests.

### DIFF
--- a/autocomplete_light/autocomplete/model.py
+++ b/autocomplete_light/autocomplete/model.py
@@ -75,8 +75,8 @@ class AutocompleteModel(object):
                 pass
 
             # Order in the user selection order when self.values is set.
-            clauses = ' '.join(['WHEN %s="%s" THEN %s' % (pk_name, pk, i) for i, pk in
-                    enumerate(self.values)])
+            clauses = ' '.join(['WHEN %s="%s" THEN %s' % (pk_name, pk, i)
+                for i, pk in enumerate(self.values)])
             ordering = 'CASE %s END' % clauses
             return choices.extra(
                 select={'ordering': ordering}, order_by=('ordering',))
--- a/autocomplete_light/autocomplete/model.py
+++ b/autocomplete_light/autocomplete/model.py
@@ -66,6 +66,9 @@ class AutocompleteModel(object):
         """
         Order choices using :py:attr:`order_by` option if it is set.
         """
+        if isinstance(self.order_by, six.string_types):
+            self.order_by = (self.order_by,)
+
         if self.values:
             pk_name = "id"
             try:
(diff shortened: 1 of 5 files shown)


autocomplete_light/autocomplete/model.py
```diff
@@ -67,9 +67,16 @@ def order_choices(self, choices):
...
+                pass
+
             # Order in the user selection order when self.values is set.
-            clauses = ' '.join(['WHEN id=%s THEN %s' % (pk, i) for i, pk in
-                    enumerate(self.values)])
+            clauses = ' '.join(['WHEN %s="%s" THEN %s' % (pk_name, pk, i)
+                for i, pk in enumerate(self.values)])
             ordering = 'CASE %s END' % clauses
             return choices.extra(
                 select={'ordering': ordering}, order_by=('ordering',))
```
rhunwicks: I'm still not clear on whether or not having `self.values` means the queryset filter won't return anything else. If the queryset will usually return ['A', 'B', 'C', 'D', 'E'] because we have `order_by = 'name'`, and then we have `self.values = ['D', 'B'] I would expect the queryset to return ['D', 'B', A', 'C', "E']. That means that we need to respect the original ordering too. E.g.:

```python
return choices.extra(
                 select={'ordering': ordering}, order_by=('ordering',)+tuple(self.order_by))
```

jpic: Right, you're absolutely correct. Fixing ...

rhunwicks:
Looks good to me. Thanks

jpic:
Awesome, thanks for your review @rhunwicks, hope to see your feedback again ;)

jmacul2:
This doesn't appear to work for me.  Double quotes around values in postgres represent column names.  For example, the above code generates this SQL:  SELECT (CASE WHEN mac="20:68:9d:8a:57:03" THEN 0 END) AS "ordering"....

If I swap the double and single quotes in the .join() that fixes it for me.

rhunwicks:
@jmacul2 - That's a good point.

Upon further consideration, I'm not sure that swapping the double quotes for single ones will be sufficient either - if you quote the value but it is actually an integer key rather than a varchar one, I think Postgresql might complain about mismatched data types and CASTing.

I don't have a dev environment for this project set up so I can't run the tests, but we should make sure we have a test for both integer and character keys where we have values in the autocomplete when we call this method.

We might need to check the datatype first and then selectively quote it. We might also be able to pass parameters to the `extra()` clause.

Also, the current code assumes that the column name will be the same as field name, which isn't necessarily so, particularly for subclassed models.

Therefore, something like this might work - but note that it is from my head, and I haven't tested it:

```python
        if self.values:
            # `choices` is a queryset, so we can derive the pk column name
            pk_name = choices.model._meta.pk.column

            # Order in the user selection order when self.values is set.
            clauses = []
            params = []
            for i, pk in enumerate(self.values)]:
                clauses.append('WHEN %s=%s THEN %s')
                params += [pk_name, pk, i]
                
            ordering = 'CASE %s END' % ' '.join(clauses)

            _order_by = ('ordering',)
            if self.order_by:
                _order_by += self.order_by

            return choices.extra(
                select={'ordering': ordering},
                order_by=_order_by,
                select_params=params)
```

jpic:
Could you guys please [try with 2.2.0rc1](http://blog.yourlabs.org/post/119860038918/django-autocomplete-light-2-2-0rc1-release-notes) ?

Thanks !

PR #419, opened 2015-05-19
ANSI SQL does not allow double quotes for enclosing string literals.

sbaum:
PostgreSQL will interpret double quoted strings as column names, so we must use single quotes instead.


jpic:
Please see #422 for followup.

jpic:
Could you try with 2.2.0rc1 which was [released yesterday](http://blog.yourlabs.org/post/119860038918/django-autocomplete-light-2-2-0rc1-release-notes) ? 

Thanks.

sbaum:
Thanks for including my fix. 2.2.0rc1 is working as expected.

PR #421, opened 2015-05-23
Absolute CSS positioning like django's calendar widget

jpic:


blueyed:
Without looking too close into this: :+1: - great stuff! :)

Issue #431, opened 2015-05-29
Absolute positioning and resize


lucky-user:
This issue is related to the choices positioning and adaptive design (like Bootstrap). 
Because of absolute positioning, introduced in 61e54ee the choices are not properly rendered. 
If you resize the browser or scroll window while displaying choices all the elements change their positions but choices don't.

jpic:
So it should be fine if we call fixPosition on browser scroll or resize ?

It's a bit annoying I recon, but we've got to go through that to have sufficient material to convince the django dev community that the admin should move from overflow: hidden to using .clearfix to clear floats.

jpic:
BTW could you make a PR for your example if you'd like to help us support this on the long run ?

lucky-user:
I think it could work but can't say for sure, as I don't understand your event handling quite right to do it.
I made PR with my example app.

jpic:
Could you try with 2.2.0rc5 please ?

Thanks !

lucky-user:
2.2.0rc5 works ok with modal dialog but it breaks with common autocomplete input because it sets this.container equals to autocomplete input itself

jpic:
Oops ! How do you suggest we improve that ? After all, we could just let
users specify the container manually but it would be better to try to
figure it automatically of course.

lucky-user:
Sorry, I didn't notice that with modal 2.2.0rc5 choices don't respond to resizes and scrolls just like before, so it doesn't help at all.
Maybe solution with resize and scroll event handling you mentioned before will do better.

jpic:
What browsers do you have that won't make the input loose focus, causing
the autocomplete to hide and that problem go away ?

For scrolling, I think it's a lot better to just start from the right
container instead of re calculating the position from the wrong container
over and over again.

About that other way, there should be a branch in my fork with an example,
but I didn't like it so much.

If we can't find a right way to support this then it would make sense to
just hide the autocomplete since the user is probably not using it anymore
when resizing or scrolling what do you think ?

lucky-user:
I have Firefox 37.0.1 and Chromium 41.0.2272.76  on Linux Mint 17. 
I finally found out that rc5 solves the problem with modal dialog scrolling so we should fix just resizing and self-container problems. 
What if we just use input's direct parent as a container or it'll break something? 
[...]

PR #433, opened 2015-05-29
Dynamic evaluation of the autocomp box container.

jpic:
In 2.2.0 series, we want most complete support of Django admin - which means
all edge cases - and it's oldschool overflow: hidden way of clearing floats.

On the other hand, we also want top-notch support for modern CSS patterns such
as the .clearfix which is the most-standard way of clearing floats IMHO and is
of course the way all modern CSS frameworks clear their floats.

By default, this traverses this.input's parents to find the nearest parent
with an 'absolute' or 'fixed' position. This prevents scrolling issues that
we'd have by boldly using <body>.


jpic:
This should fix issues #431 and make the example app in #432 work properly.

PR #436, opened 2015-05-31
Removed duplicate test: test_input_disappears does the same

jpic:


blueyed:
Have you checked that reverting https://github.com/yourlabs/django-autocomplete-light/pull/334/files would be caught by `test_input_disappears`?

jpic:
@blueyed 
That commit is not there anymore because we're back into absolute positioning.

One good news is that I've had BrowserStack support by email, and they said they'd let us test our Open Source repos with it for free so that's definitely going to help and be fun to learn about ;)

Issue #62, opened 2012-11-15
ValueError: "invalid literal for int() with base 10" when submitting an empty form POST


migajek:
Hi,
I've started using autocomplete-light today, for autocompletion of ForeignKey on my custom ModelForm.
however when submitting an empty POST request, mentioned in title ValueError is being thrown, pointing to the template ("widget.html") line:
 {% for choice in autocomplete.choices_for_values %}

which leads to the following:
autocomplete/model.py in choices_for_values:

return self.order_choices(self.choices.filter(
            pk__in=self.values or []))

I've debugged that line and believe the problem is that, when empty POST is submitted, self.values contains one value, an EMPTY UNICODE STRING (that is, self.values == [u''])

I don't know autocomplete_light internals well enough to find out why is that

blueyed:
This should be fixed (the PR was merged).
Probably re-opened accidentally.

PR #450, opened 2015-06-23
Foreign key with 0 as PK is valid.

Velmont:
If you added an FK with 0 as primary key, autocomplete would complain
that it was invalid because it filtered out everything Python considered
falsy after a fix to #62 was provided in #298.

After this, autocomplete validation will check directly for the empty
string '' instead of falsy values.

### DIFF
--- a/autocomplete_light/example_apps/autocomplete_test_case_app/models.py
+++ b/autocomplete_light/example_apps/autocomplete_test_case_app/models.py
@@ -20,6 +20,10 @@ class NonIntegerPk(models.Model):
                                    related_name='inline')
 
 
+class CustomIntegerPk(models.Model):
+    id = models.IntegerField(primary_key=True)
+
+
 class SubGroup(Group):
     pass
 
--- a/autocomplete_light/autocomplete/model.py
+++ b/autocomplete_light/autocomplete/model.py
@@ -99,7 +99,7 @@ class AutocompleteModel(object):
         """
         assert self.choices is not None, 'choices should be a queryset'
         return self.order_choices(self.choices.filter(
-            pk__in=[x for x in self.values if x]))
+            pk__in=[x for x in self.values if x != '']))
 
     def choices_for_request(self):
         """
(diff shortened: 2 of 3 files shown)


Velmont:
Shait. Stupid MySQL. So, any ideas on how to make that test not run for MySQL?

Velmont:
OK, I rewrote the test to use its own model.  Some tests expected not having to fill in the ID field, so I overwrote those.

I thought about doing some helper or other way of repeating the code, but in the end went with a simple overwrite.  I don't know what you prefer.  I went for the more explicit version instead of DRY.  If you have a special way you want it, I'll ofc change it :)

jpic:
Thanks a heap, but I really prefered the earlier version where you added a very light test, wouldn't it work to add it back and try to find another way to skip the test on MySQL since it doesn't make sense ?

Velmont:
Actually MySQL can work with zero-based fk's too. Only, it can't work with zero in AUTO_INCREMENT primary keys. :)

So this way actually checks that it works with MySQL too.

Though I can check and skip MySQL using the simpler test I did before. django.db.connection.vendor == 'mysql' ought'a do it.

autocomplete_light/example_apps/autocomplete_test_case_app/models.py
```diff
@@ -20,6 +20,16 @@ class NonIntegerPk(models.Model):
...
 
 
+class CustomIntegerPk(models.Model):
+    id = models.IntegerField(primary_key=True)
+    name = models.CharField(max_length=10)
+    relation = models.ForeignKey('self', null=True, blank=True)
+    noise = models.ForeignKey('basic.FkModel', null=True, blank=True)
+
+    for_inline = models.ForeignKey('self', null=True, blank=True,
+                                   related_name='inline')
```
jpic: Could you remove things which are tested in other models here please ?

jpic:
Could we have a test in tests/autocomplete/test_model.py's AutocompleteModelTestCase rather ? I don't think we need such a high level test to prevent regressions on this, a unit test will be faster and do the same ;)

Velmont:
Hehe.  I was a bit afraid something else would break it later :P  But yeah, it does not give a lot extra.

I think I found a good place for it now.  Didn't remove the added model, but made it much less imposing since it's now only testing the core of zero primary key.

jpic:
As good as I could pray for, merging. Please let us know if you need any emergency release.

Velmont:
Thank you :)  No, I have it patched locally so far.

PR #456, opened 2015-07-06
tox.ini: add DISPLAY to passenv

blueyed:
This is required to fix the following error with `tox -e py34-django18`
(using Firefox):

> selenium.common.exceptions.WebDriverException: Message: The browser appears to have exited before we could connect.

@jpic 
Did this work for you locally?
Are you using another webdriver?


jpic:
Fixes the problem for me, nice one B)

Issue #452, opened 2015-07-01
Inline autocomplete breaking 


nwaxiomatic:
I have no real idea why this would be breaking, I am just using a simple autocomplete in an inline and getting this error

```
missing FROM-clause entry for table "subjects_subject"
LINE 1: SELECT (CASE WHEN Subjects_subject.id='1' THEN 0 END) AS "or...
```

any help is appreciate, here is traceback

Environment:

```
Request Method: POST
Request URL: http://104.236.221.79/admin/Tutor/tutorstudentrelationship/12/

Django Version: 1.8.2
Python Version: 2.7.6
Installed Applications:
('autocomplete_light',
 'favicon',
 'wpadmin',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.sites',
 'apptemplates',
 'jquery',
 'crispy_forms',
 'bootstrap3',
 'localflavor',
 'datetimewidget',
 'fm',
 'django_fsm',
 'fsm_admin',
 'djwkhtmltopdf',
 'schedule',
 'easymoney',
 'phonenumber_field',
 'django_countries',
 'postal',
 'babel',
 'easy_thumbnails',
 'timezone_field',
 'allauth',
 'allauth.account',
 'allauth.socialaccount',
 'Accounting',
 'Billing',
 'GroupInvitations',
 'BrownstoneTutors',
 'Sessions',
 'Student',
 'Tutor',
 'Client',
 'Subjects')
Installed Middleware:
('django.middleware.common.CommonMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'django.middleware.locale.LocaleMiddleware')


Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/dist-packages/django/contrib/admin/options.py" in wrapper
  616.                 return self.admin_site.admin_view(view)(*args, **kwargs)
[...]

PR #464, opened 2015-07-16
#452 Fix the case where there is a cap in the pg table or column name

jpic:


jpic:
Wish I had anything better, @blueyed could you review this and eventually merge it if you see fit ?

PR #477, opened 2015-08-17
Widget's render should use attrs even when created wthing __init__

sandroden:
While working with autocomplete I realized that any attribute added from within __init__ will not be used by the render function. I thing this is not optimal as it is counterintuitive and results in code that is not DRY.

TIA
sandro
*:-)


jpic:
Very good, thanks !

jpic:
All tests passed, merging, thanks a heap !

PR #491, opened 2015-09-16
Explain some basic concepts in tutorial

jonashaag:
This isn't explained anywhere in the documentation, but it's the very basics of how autocomplete-light works on the surface, so I figured this is a valuable addition to the tutorial.

Maybe we should also add the full ``Person`` and ``Order`` definitions somewhere or even add the Person/Order app to the example apps in the repository.


jpic:
Awesome, thanks !

PR #494, opened 2015-09-17
ModelChoiceFields: Watch changes to 'queryset'

jonashaag:
A ModelChoiceField's 'queryset' attribute may be overwritten after the field has been created, for example in order to limit the choices that are offered. We should also reflect these changes in the autocompletion.

A common idiom is to override ``queryset`` in a view or form, like in this example: https://docs.djangoproject.com/en/dev/ref/forms/fields/#fields-which-handle-relationships. Currently, these overrides are not reflected in the autocomplete. Hence it is possible to select related objects in the autocomplete that aren't part of ``queryset`` anymore, resulting in a validation error.

This patch fixes this by always reflecting changes to ``queryset`` in the autocomplete's choices.

### DIFF
--- a/autocomplete_light/fields.py
+++ b/autocomplete_light/fields.py
@@ -86,16 +86,30 @@ class MultipleChoiceField(ChoiceField, forms.MultipleChoiceField):
     widget = MultipleChoiceWidget
 
 
-class ModelChoiceField(FieldBase, forms.ModelChoiceField):
+class ModelChoiceFieldBase(FieldBase):
+    def _get_queryset(self):
+        return self._queryset
+
+    def _set_queryset(self, queryset):
+        self._queryset = queryset
+        self.widget.choices = self.choices
+
+        # Also update autocomplete choices
+        self.autocomplete.choices = queryset
+
+    queryset = property(_get_queryset, _set_queryset)
+
+
+class ModelChoiceField(ModelChoiceFieldBase, forms.ModelChoiceField):
     widget = ChoiceWidget
 
 
-class ModelMultipleChoiceField(FieldBase,
+class ModelMultipleChoiceField(ModelChoiceFieldBase,
         forms.ModelMultipleChoiceField):
     widget = MultipleChoiceWidget
 
 
-class GenericModelChoiceField(FieldBase, forms.Field):
+class GenericModelChoiceField(ModelChoiceFieldBase, forms.Field):
     """
     Simple form field that converts strings to models.
     """
--- a/autocomplete_light/tests/test_forms.py
+++ b/autocomplete_light/tests/test_forms.py
@@ -172,8 +172,10 @@ class ModelFormBaseMixin(BaseModelFormMixin):
         class ModelForm(autocomplete_light.ModelForm):
             class Meta(DjangoCompatMeta):
                 model = self.model_class
-                widgets = {'relation': self.widget_class(widget_attrs={
-                    'class': 'test-class', 'data-foo': 'bar'})}
+                widgets = {'relation': self.widget_class(
+                    autocomplete=self.autocomplete_name,
+                    widget_attrs={'class': 'test-class', 'data-foo': 'bar'}
+                )}
 
         self.form = ModelForm()
 


jpic:
Thanks ! But we've got to find a backward compatible way.

jonashaag:
One issues seems to be that ``choices`` is overridden when it's a list of querysets. In which cases does this happen? Where can I find the code that creates the list?

jpic:
That would be generic fk/m2m autocompletes

jonashaag:
There seems to be a bug in the way autocompletes are found for widgets. If a widget is initialized with `autocomplete=None` (the default case) and a generic autocomplete has been registered earlier on, it is used as the widget's autocomplete even though they're not related in any way.

```py
    def get_autocomplete_from_arg(self, arg=None):
        ...
        elif arg is None:
            return self.default_generic
        else:
            raise AutocompleteArgNotUnderstood(arg, self)
```

Why is this case there in the first place? Why would we default to some unrelated generic autocomplete?

jonashaag:
bump

jpic:
We'll need tests to merge this ;)

jonashaag:
@jpic Could you try to answer my question in the bug report above? https://github.com/yourlabs/django-autocomplete-light/pull/494#issuecomment-141775417

jpic:
Right, I still had a draft reply in my mailbox, sorry about that, but it's a feature, not a bug :)

What do you think should happen if a widget got instanciated without autocomplete ?

jonashaag:
> What do you think should happen if a widget got instanciated without autocomplete ?

Not sure, maybe don't display any autocomplete at all? Why would you want to see Generic Foreign Key autocompletes for some completely unrelated field?

jpic:
I see, so we'd have to implement logic in the autocomplete widget to not be an autocomplete widget if the autocomplete is wrong ? But shouldn't the user then just **not** use an autocomplete widget where (s)he doesn't want to have an autocomplete widget ? There's got to be a better way and I sort of agree with you that think feature might be confusing, maybe it's implemented wrong or something.

Removing the default-generic feature means breaking backward compatibility:
- changing [test_default_generic](https://github.com/yourlabs/django-autocomplete-light/blob/master/autocomplete_light/tests/test_registry.py#L152-L154)
- issue a middle version increment release, since it's a BC break in an undocumented feature, it shouldn't affect so much users,

My prefered options here are:
- fixing whatever bug you found in this feature to make it better,
- raise an exception instead,
- implement a system check to prevent the user from having widgets without autocompletes

jpic:
Wait, shouldn't we talk about this on another topic ? I completely lost track about this code change proposal. I'd like to merge this but we'd need to cover this code with tests.

jonashaag:
This should now pass the test suite. The Travis build fails for some weird reasons.

For the default generic business have a look at my `remove-default-generic` branch. If there's anything to discuss on this topic maybe open a new PR as soon as this one is merged. (`remove-defaulte-generic` depends on this PR)

jpic:
Could you please try rebasing on master ?

Thanks

jonashaag:
It's already rebased.

jpic:
Tests are passing right now, I'm sorry but could you rebase again and squash both commits into one ?

Thanks a heap !!

jpic:
I really don't want to do another release without your patch in there ^^

jpic:
Wait before rebasing, we still need a regression test.

jonashaag:
I've since switched away from django-autocomplete-light; I've had too many other issues. Not interested in working on this anymore.

jpic:
Yeah I'm sorry, I don't have as much time as I used to. Thanks for your contribution, it'll be in the next release for sure.

jpic:
Closed with 181e68d

jpic:
Oops, I think I'll have to revert this, you can't change the autocomplete in a form instance, because the change would not be applied when the autocomplete is instanciated by [AutocompleteView](https://github.com/yourlabs/django-autocomplete-light/blob/master/autocomplete_light/views.py#L59). No wonder why you've had issues doing this, but I don't think you can do it with any autocomplete app unless maybe by using the `request.session` object and all.... You queryset change would apply only in `choices_for_values` and not `choices_for_request`.

Reverting.

jpic:
BTW, the test was easy to add:

```
diff --git a/autocomplete_light/tests/test_fields.py b/autocomplete_light/tests/test_fields.$
index e336224..4d2afb9 100644
--- a/autocomplete_light/tests/test_fields.py
+++ b/autocomplete_light/tests/test_fields.py
@@ -73,6 +73,13 @@ class ModelChoiceFieldTestCase(BaseMixin, TestCase):
         self.assertEqual(list(test.choices),
                          [('', '---------'), (1, 'public'), (3, 'linked')])

+    def test_change_queryset(self):
+        test = self.field_class(self.TestAutocomplete, required=True)
+
+        test.queryset = test.queryset.filter(pk=None)
+
+        assert test.queryset == test.autocomplete.choices
+
```

jpic:
My bad, we can keep this, it works, BUUUUUUUUUT you HAVE to reproduce the change in `choices_for_request()`. Ie for:

    field.queryset = field.queryset.filter(owner=user)

You MUST have:

    def choices_for_request(self):
        return self.choices.filter(owner=self.request.user)

So, that's really tricky ;)

PR #497, opened 2015-09-23
support to register model by 'app_name.ModelName' or

luzfcb:
'full.path.to.Model'

My use case:

```python
import autocomplete_light
from django.conf import settings

USER_MODEL = getattr(settings, 'AUTH_USER_MODEL', 'auth.User')


class UserAutocomplete(autocomplete_light.AutocompleteModelBase):
    search_fields = ['^first_name', 'last_name']

    def choice_label(self, choice):
        return choice.get_full_name()


autocomplete_light.register(USER_MODEL, UserAutocomplete)

```
I could not maintain compatibility with django <1.6. I accept suggestions for how to do this

### DIFF
--- a/autocomplete_light/compat.py
+++ b/autocomplete_light/compat.py
@@ -9,7 +9,51 @@ else:
     patterns = None
 
 
+try:
+    # Django 1.7 or over use the new application loading system
+    from django.apps import apps
+    get_model = apps.get_model
+except ImportError:
+    from django.db.models.loading import get_model
+
+try:
+    from django.utils.module_loading import import_string
+except ImportError:
+    from importlib import import_module
+    from django.utils import six
+    import sys
+
+    def import_string(dotted_path):
+        """
+        Import a dotted module path and return the attribute/class designated
+        by the last name in the path. Raise ImportError if the import failed.
+        Backported from Django 1.8
+        """
+        try:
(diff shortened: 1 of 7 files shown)


jpic:
Did you consider using django-swappable-models ?​

jpic:
We're going to drop compatibility for django < 1.8 in 2.3 release in january, we'll be able to merge this.

jpic:
Hi @luzfcb !

Could you please rebase on origin/master ? 

PR #555 deprecates earlier django versions, so we should be able to merge that one soonish.

Thanksss !

luzfcb:
@jpic Done.

autocomplete_light/compat.py
```diff
@@ -9,7 +9,51 @@
...
     if patterns:
         return patterns('', *urls)
     return urls
+
+__all__ = (
+    'get_model',
+    'import_string',
+    'urls',
+    'url',
+    'patterns'
```
luzfcb: I had added `urls`, `url`, `patterns`, now I'm thinking it would only be necessary export `urls` function (and `get_model`, `import_string`)

jpic:
You rock, I'm currently in the process of merging, thankssss !!!

jpic:
Released in 2.3.0

PR #500, opened 2015-10-02
Fixes an import error for Django 1.9a

arnists:


arnists:
I've seen the Travis CI logs and I don't understand how this fails.  Is comparing VERSION the way I do it in the PR not kosher?

arnists:
Neverr mind, I understand where I failed.

PR #501, opened 2015-10-02
Django19fix

arnists:


jonashaag:
Note that `patterns` has *not* been removed in Django 1.9 but in Django 1.10, so this ticket should be called "Django110 fix".

blueyed:
@jonashaag 
But there's a deprecation warning then that gets fixed with 1.9, right?

@arnists 
Can you squash your commits into a single one, please?

arnists:
I don't understand why the build fails.  I just squashed the commits as asked to, changes are the same.  Were any other commits accepted onto main with failing builds?

jonashaag:
> Were any other commits accepted onto main with failing builds?

Nope, build seems ok, but you could try by opening a pull request with an empty commit.

blueyed:
Tests failures seem to be caused by a bug in pytest-django: https://github.com/pytest-dev/pytest-django/issues/280

PR #504, opened 2015-10-06
tests reqs: pin pytest-django==2.9.1

blueyed:
Could be unpinned again, but it makes sense to not break tests because
of that in the future.

If you agree, other deps could be pinned also?!

### DIFF
--- a/test_project/test_requirements_without_django.txt
+++ b/test_project/test_requirements_without_django.txt
@@ -5,4 +5,4 @@ selenium
 lxml==3.4.4
 cssselect
 pytest
-pytest-django==2.8.0
+pytest-django==2.9.1


jpic:
It would definitely help to pin versions ! Thanks @blueyed

blueyed:
You're very welcome!
See https://github.com/pytest-dev/pytest-django/issues/282 for the new issue with 2.9.1.

jpic:
Could you try to rebase on master please ? I think tests will pass then.

blueyed:
Nope, still failing.

jpic:
I don't think we need this anymore, do we ?

blueyed:
> I don't think we need this anymore, do we ?

What do you mean?  This PR is about using the latest pytest-django again.

I've re-pushed it rebased on master.

blueyed:
Seems to be still this issue: https://github.com/pytest-dev/pytest-django/issues/282

PR #508, opened 2015-10-07
Py2.7: use new style classes

jpic:
Else, it fails as such::

    self = <autocomplete_light.tests.test_dependent.DependentAutocompleteEmptyFormTestCase testMethod=test_alpes_in_france>

        def set_implicit_wait(self):
    >       self.selenium.implicitly_wait(WAIT_TIME)
    E       AttributeError: Selenium instance has no attribute 'implicitly_wait'


blueyed:
:+1:

PR #515, opened 2015-10-26
Update polish translation

mpasternak:
Business as usual.


jpic:
I have no idea why travis marked it as failed, all jobs passed ^^

Thanks ;)

PR #516, opened 2015-10-27
Corrected french translation for "widgets.py:82"

apinsard:
- "votre recherche" instead of "votre rechercher".
- "dans cette autocomplétion" does not sound really goud. It would be better to say "dans ce champ de saisie automatique". However "dans ce champ" sufficient and sounds better.


jpic:
Merci frêre !

PR #521, opened 2015-11-15
Host demo on openshift

jpic:


jpic:
Need to close the PR for travis to deploy on openshift.

PR #522, opened 2015-11-15
Compat accross django versions in tests

jpic:


jpic:
The point here is for tests to pass on django 1.10 (tested as "djangomaster").

PR #534, opened 2015-12-02
Make sed command OS X/BSD-compatible

mfcovington:
The `sed` command in the README for upgrading projects to Django 1.9 does not work on OS X and gives the following error, for example:

    sed: 1: "./__init__.py": invalid command code .

This is because [OS X uses BSD `sed` which either requires `''` after the `-i` if there is no backup file extension](http://stackoverflow.com/a/16746032/996114) ~~or the use of `-e`. I used `-e` in the PR, because I think it looks cleaner than having extra quotes, but the following change also works:~~

    find . -name '*py' | xargs sed -i '' 's/import autocomplete_light/from autocomplete_light import shortcuts as autocomplete_light/'

**Update:** I've actually switched the PR to use the `... sed -i '' ...` approach, because the original approach I used created unintended backup files with a suffix of `-e`.

### DIFF
--- a/README
+++ b/README
@@ -20,7 +20,7 @@ DAL has been ready for Django 1.9 since April 2015 thanks to @blueyed & @jpic.
 **HOWEVER** due to the app loading refactor in 1.9 you should apply the
 following::
 
-    find . -name '*py' | xargs sed -i 's/import autocomplete_light/from autocomplete_light import shortcuts as autocomplete_light/'
+    find . -name '*.py' | xargs perl -i -pe 's/import autocomplete_light/from autocomplete_light import shortcuts as autocomplete_light/'
 
 See the test_project running on Django 1.9 and its new cool admin theme:
 http://dal-yourlabs.rhcloud.com/admin (test:test).


blueyed:
That fails with GNU's sed:

> sed: can't read s/import autocomplete_light/from autocomplete_light import shortcuts as autocomplete_light/: No such file or directory

Does it work with `-i''` for you? (with the space removed)

jpic:
Could you maybe add that command to the docs rather than replacing the
existing one ?

Pretty cool, thanks for sharing !!

blueyed:
Apparently not: http://stackoverflow.com/a/5694430/15690

blueyed:
So I'd say to add a note for OSX/BSD users instead.

jpic:
Yeah that's what i meant, add instead of replace

mfcovington:
Which would you prefer:

    # For GNU sed
    find . -name '*py' | xargs sed -i 's/import autocomplete_light/from autocomplete_light import shortcuts as autocomplete_light/'

    # For OSX/BSD sed
    find . -name '*py' | xargs sed -i '' 's/import autocomplete_light/from autocomplete_light import shortcuts as autocomplete_light/'

Or this:

    find . -name '*.xx' | xargs perl -i -pe 's/import autocomplete_light/from autocomplete_light import shortcuts as autocomplete_light/'

blueyed:
The shorter one.  It should not bloat the documentation, and instructions are clear / can be adopted.

blueyed:
Please force-push it as a single commit then.

mfcovington:
Done. Thanks.

blueyed:
Thank you!

PR #539, opened 2015-12-15
Make the JS Urls in the Examples work with https too

lazerscience:
If you copy and paste the example code it will fail when using https for the connection.


blueyed:
Should be `https://` instead, see http://www.paulirish.com/2010/the-protocol-relative-url/.

jpic:
Can we have https:// instead ?

jpic:
That's good enough for now.

PR #541, opened 2015-12-16
Use error_messages on FieldBase to allow overrides

dsanders11:
This PR lets you override the error message for failure to validate by passing error_messages when instantiating fields such as:

```
foo = autocomplete_light.fields.ChoiceField(autocomplete="FooAutocomplete", label="Foo", error_messages={'invalid': "Check your foo"})
```

I have not tested it, but it should also enable:

```
class BarModelForm(autocomplete_light.ModelForm):
    class Meta:
        model = Bar
        error_messages = {
            'foo': {
                'invalid': "Check your foo"
            }
        }
```

### DIFF
--- a/autocomplete_light/fields.py
+++ b/autocomplete_light/fields.py
@@ -14,6 +14,10 @@ __all__ = ['FieldBase', 'ChoiceField', 'MultipleChoiceField',
 
 
 class FieldBase(object):
+    default_error_messages = {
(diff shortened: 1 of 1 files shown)


jpic:
Pretty cool ! We need a regression test though

dsanders11:
@jpic, I probably won't be able to add a test for a few weeks, pretty swamped with other work. I'll eventually get around to it though, if someone doesn't beat me to it.

PR #542, opened 2015-12-20
Official request object support in validation

jpic:


jpic:
I merged a mistake in the current master, which this PR reverts, and adds a proper fix for. Review by @jonashaag would be appreciated.

jonashaag:
:-1: this reverts 3cbc4dc92d3bc902d11e3b29a7f9ea9cc3f8dcb7

jpic:
Yep, but it's the right way to address the use case. Else you have to duplicate your security logic in choices_for_request and field.clean() which is error prone and dangerous in the case of security.

jonashaag:
> Took me quite some time to get this right, it's perhaps my 6th or 7th session on this ^^

Maybe autocomplete-light is too complex and some larger refactoring is necessary to make things simple to understand again. That was my impression too when working on commit/3cbc4dc92d3bc902d11e3b29a7f9ea9cc3f8dcb7.

jpic:
Ok, thanks for your productive feedback !

jpic:
If a [5 SLOC change](https://github.com/yourlabs/django-autocomplete-light/commit/04c7f45a108f616a0e7a54db786b94e3903f0ff7#diff-dfa9735d495c735fbacef4fadb637a21R71) was too complex for me, then I should definitely find another job !!!

Note how my version changes less code than yours, is more predictible, not to mention that it's tested and documented :)

jpic:
I think you misplaced the complexity here, what made this hard is not DAL, it's that Django's form field validation process completely ignores the request object, so maybe when you really try to really understand what's happening here you'll change your opinion ;)

jonashaag:
One of the sources of complexity is that the `choices` attribute may be set in many cases, under various circumstances. It even isn't clear if it's going to be changed at some other place later on if set at some place. It hijacked `__getattr__` to look for `"choices"` changes to find the places it is set from. This definitely **is** complexity, namely the complexity of having multiple places of definition of a value.

jpic:
If you're changing choices from outside the Autocomplete class, then your change will not apply in the AutocompleView and you'd be sabotaging your autocompletes - that may be one of the reason that makes you think that it "doesn't work".

When you understand this, you understand why choices should be changed only from inside the Autocomplete class, which is used by both the form and the view.

How would changing choices in the Field or Widget apply in the Autocomplete view which are instanciated from completely different requests thus threads ?

jpic:
You know what, try to find a better design than having a central Autocomplete class which encapsulates business and rendering logic for the field (for validation), widget (for rendering) and view (for choice selection) and I'll send you over 50 bucks, here's your challenge if you accept it: https://github.com/jpic/django-autocomplete-compare

jonashaag:
Wow. First off, I'm very sorry if any comment I made felt like a personal insult to you. I never intended that. Secondly, I must have confused the `choices` and `autocomplete` attributes above. `choices` isn't set from many places, but `autocomplete` is, and it is `autocomplete` that took me quite some debugging to understand where it was coming from.

However, I still think the overall architecture of autocomplete-light is hard to grasp, or maybe just not explained very well. The documentation contains a lot of concrete examples and case studies, but what I think is missing is bird's eye view on the concepts and how they relate which each other. f6cba54657452e34f6bfbd97f5ed93a1e2c63f9f is a very handwavy first step in this direction.

As for this business logic vs. validation issue: I don't get it. Maybe you could point me to a very simple example that explains the problem. For now I don't see why you can't put the validation logic into your form.

jpic:
It's not that you can't, it's that it's not enough.

jpic:
Pointing you to a very simple example ? What did I just do omfg https://github.com/jpic/django-autocomplete-compare

jpic:
Look, I'm closing this as counter productive, this PR as docs, tests, and I also pointed you to a simple, isolated example in https://github.com/jpic/django-autocomplete-compare, but apparently, the only thing you've reviewed is that your hazardous, insecure commit was reverted, I'm moderating this set of comment as counter productive. Thanks a lot for your time.

jpic:
Plan has changed, see 2.3.0 release.

jonashaag:
This comes as a surprise, what made you change your mind? Didn't the approach you suggested work out? It didn't look wrong to me…

jpic:
I don't mind leaving it as such in v2, as I'm working on v3 already:

I think the best way is to store the queryset somehow in the session, so
that you could change the queryset from outside (ie. in the view) and have
changes persisted in the autocomplete view natively.

At the same time, I'm trying to make widgets / field classes inherit from
Autocomplete classes so that we wouldn't have to pass around manually in
the case of ModelChoiceField (let django do it).

jpic:
Here's a more detailed paper about how v3 - which completely eliminates this problem - looks like, if you're curious:

https://github.com/yourlabs/django-autocomplete-light/blob/v3/README

As you can see, v3 is vastly superior to v2 while having a much, much smaller codebase (no view, no registry, no autocomplete class).

PR #547, opened 2016-01-10
Check that genericm2m is in INSTALLED_APPS

jpic:
This prevents loading genericm2m if it's just installed in the
environment.


jpic:
Released in 2.3.0

PR #548, opened 2016-01-10
Fix #493: capitalize() is not cross-language

### LINKED ISSUE #493: Autocomplete "no matches found" lowercased
In the German translation, the "no matches found" text seems to be lowercased:

screen shot 2015-09-16 at 13 43 27

This should read "Keine Treffer gefunden", like it's in the localization file.

jpic:


jonashaag:
:+1:

PR #558, opened 2016-01-21
v3: The autocomplete app where YOU are the hero

jpic:


jpic:
This PR contains a PoC for the new design (using select2), please refer to [README](https://github.com/yourlabs/django-autocomplete-light/blob/v3/README) for specs.

jpic:
Note that I'm using [django-querysetsequence](https://github.com/percipient/django-querysetsequence) for generic autocompletes now and the API is soooooo sexy, since it's compatible-enough with the QuerySet API, that means even more code removal !

jpic:
So there's indeed a need for a registry, particularely when it comes to generic relations :)

This is what the new one looks like:

```
class WidgetMixin(object):
    def __init__(self, url=None, *args, **kwargs):
        self.url = url
        super(WidgetMixin, self).__init__(*args, **kwargs)
        self.attrs['data-autocomplete-light-url'] = url 

    @cached_property
    def view(self):
        return resolve(self.url).func.view_class
```

Yep, that replaces a 166 SLOCs module !

jpic:
For the curious, here's what the feature list looks like:

django-autocomplete-light v3: the autocomplete app where you are the hero
                              so simple it should hang itself with a wifi cable.

FEATURES

- Django (multiple) choice support,
- Django (multiple) model choice support,
- Django generic foreign key support,
- Django generic many to many relation support,
- External app supported in extensions, ie. django-taggit,
- Multiple widget support: select2.js, easy to add more.
- Creating choices that don't exist in the autocomplete,
- Offering choices that depend on other fields in the form,
- Dynamic widget creation (ie. inlines), supports YOUR custom scripts too, in case you need any
- Public and clean test API for your awesome autocompletes, to support YOUR custom
  use cases too,
- Django-autocomplete-light V2 compatibil-ish extension in dal_smart, for those
  who like DRY and already have a solid understanding of django and autocompletes.

CHANGES from DAL v2

- no registry,
- no magic,
- no url / view,
- multiple autocomplete script support
- use-case based test system, see tests/ directory **at the root of the repo**,
- no autocomplete classes, but dal_smart extension provides DumbField and
  DumbWiget, which delegate rendering, validation and all to SmartView for DRY
  purposes, SmartView reuses the API from Autocomplete because it has just kept
  on checkmating me no matter how many nights and days I try every possible
  refactoring combination. But it comes at a price: pass the request object to
  the field or we can't garantee your security :)

jpic:
Features left to implement / test / demo:

- autocompletion filtered per request user
- autocompletion filtered by current form state
- tags (django-taggit, django-taggit-ng, django-tagging)

Note that in addition to django-generic-m2m, support for django-gm2m was implemented.

jpic:
Intead of tight coupling like in v2, check out all the loosely coupled modules we end up with: https://github.com/yourlabs/django-autocomplete-light/tree/v3/src

jpic:
This is feature complete in the admin ! 

Test coverage, is a lot better than I thought it would be (97%).

An interresting change is that V3 relies on forward compatibility with django ModelForm, properly delegating business logic to form fields, instead of a soup of ModelForm mixins.

Also, instead of adding add-another feature directly into DAL, I'd like to add it in a separate app providing something like django admin's RelatedFieldWidgetWrapper, particularely since I've seen the code pasted in another autocomplete app's code ^^ Note that v3 has better object creation feature integration.

Dal_smart is not there yet, I don't think it's blocking the v3 release because it should provide equivalent features than v2 and we're going to maintain v2 for a while. Only, for new comers v3 is much easier to support and maintain, so we do want this out there.

So far here's the plan I thought could work:

- Don't support new versions of django in v2, we support up to 1.10 already so that means we'd support v2 until both 1.8 and 1.10 are deprecated + ~6 months or something on a best effort basis,
- Release 3.0 as soon as possible, that means after writing docstrings and documentation,
- Implement the non-admin equivalent update-forms in the example apps, with several dynamic inline formset support (the new JS was designed to support any widget as you know, but also on any kind of dynamic inline script),
- Release 3.1 with dal_smart, with actual DRY-style autocompletes and nice things from JAL like templated options etc ...

jpic:
It passes and I'm working on the last page of documentation. Expect a release in minutes !!!

PR #568, opened 2016-02-04
Fixed syntax error in example

Eraldo:


jpic:
Nice, thanks !

blueyed:
Thanks indeed - I've seen this being copy'n'pasted actually.. :)

PR #574, opened 2016-02-04
fix taggit example in tutorial

Athemis:
The taggit example as provided by the tutorial did not work. This pull request will restore basic functionality.

Cheers,
Alex


jpic:
Thanks a heap !!

PR #579, opened 2016-02-05
WIP - MagicForm

jpic:


jpic:
This design changes a lot from v2, like the rest.

In FutureModelForm, we already add the ability to encapsulate business logic in form fields, by adding support of new methods: one method to get the initial field value from the model, one to save the field's value in the model, and one to save relations. Currently in Django all this is only doable by the model field, and not overridable per modelform, FutureModelForm changes this.

In MagicModelFormMetaclass, we're just giving apps the possibility to register form field classes to the model form metaclass. Our modelform metaclass iterates over the model fields to have in the form, and if any of them has a registered form field, then it will call the factory() class method of the form field, if it returns a form field instance then it should be used by default.

Of course, it is then the responsibility of the form field, to iterate over the urls and figure out which url it should use by inspecting the view classes. In the form field's factory method, the form's Meta is also passed, which allows the generated form field to use Meta for configuration.

There's still some work, but I think this design makes it hackable by anybody, more easy to support and to contribute to. Let me know what you think !

src/dal_magic/forms.py
```diff
@@ -0,0 +1,72 @@
...
+
+        # Maybe the parent has a meta ?
+        if meta is None:
+            for parent in bases + type(cls).__mro__:
+                meta = getattr(parent, 'Meta', None)
+
+                if meta is not None:
+                    break
+
+        return meta
```
jpic: This bad code should be replaced by: request.js.add('dal/dist.js')

src/dal_select2/fields.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Select2ModelMultipleChoiceField(
+    MagicFieldMixin,
+    forms.ModelMultipleChoiceField):
+
+    widget = ModelSelect2Multiple
+
+    @classmethod
+    def factory(cls, model, form, field):
+        raise NotImplemented()
```
jpic: This is bad code, there should not need to invent form fields when it just wants to set the default widget.

test_project/tests/test_magic.py
```diff
@@ -0,0 +1,41 @@
...
+from dal_magic.forms import MagicForm
+
+from django.apps import apps
+from django.conf import settings
+
+
+class TestMagicForm(unittest.TestCase):
+    def test_magicform(self):
+        #for app in settings.INSTALLED_APPS:
+        for app in ['select2_foreign_key']:
```
jpic: Apparently this was not even finished.

PR #585, opened 2016-02-08
Update select2.js to support widget placeholder

kliver:
Support for placeholder value.

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -8,7 +8,7 @@
         $(this).select2({
             tags: element.attr('data-tags'),
             debug: true,
-            placeholder: '',
+            placeholder: $(this).attr('placeholder'),
             minimumInputLength: 0,
             allowClear: ! $(this).is('required'),
             ajax: {


jpic:
As you can see, this doesn't pass test. When I try it locally, I can see it breaks when the attr is "undefined". I think it would work better with `$(this).attr('placeholder') || ''`, adding an example ie. in select2_foreign_key/forms.py would be great too, and ensure that this case is also tested !

kliver:
Oh sorry, first time using this thing.

jpic:
No problem ! In this case, click "Details" on the failed test, here you'll see:

```
======================================================================
FAIL: test_can_unselect_option (select2_foreign_key.test_functional.AdminForeignKeyTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/yourlabs/django-autocomplete-light/test_project/select2_foreign_key/test_functional.py", line 68, in test_can_unselect_option
    story.assert_selection_persists('', '')
  File "/home/travis/build/yourlabs/django-autocomplete-light/src/dal/test/stories.py", line 117, in assert_selection_persists
    self.assert_selection(value, label)
  File "/home/travis/build/yourlabs/django-autocomplete-light/src/dal/test/stories.py", line 112, in assert_selection
    self.assert_label(label)
  File "/home/travis/build/yourlabs/django-autocomplete-light/src/dal/test/stories.py", line 93, in assert_label
    six.text_type(self.get_label()),
AssertionError: u'' != u'baa07b76ce1b11e5ac4e82af6c8ebe6a'
+ baa07b76ce1b11e5ac4e82af6c8ebe6a
```

Perhaps this is due to a bug in select2, but un-selecting an option doesn't work when placeholder is undefined. Could you perhaps change your patch to something like `placeholder: $(this).attr('placeholder') || ''` and see if that works better ?

Thanks !!!

jpic:
I just tried and this worked:

```

class TestForm(forms.ModelForm):
    class Meta:
        model = TestModel
        fields = ('name', 'test')
        widgets = {
            'test': autocomplete.ModelSelect2(
                url='select2_fk',
                attrs={'data-placeholder': 'Autocomplete ...'},
            )   
        }   
```

So perhaps we could just add an example in the documentation if you'd like ?

jpic:
Closing in favor of #662

PR #589, opened 2016-02-08
Smartview (replacement for Autocomplete classe)

jpic:
Autocomplete classes break Django's separation of field / widget / view in a
sort of clumsy way. This design makes it more obvious what happens being the
scene.


jpic:
Guillouf's idea is better in #995

PR #594, opened 2016-02-12
Fixed variable names in documentation.

Eraldo:
I suppose "guitars" and "trumpets" was used before using "countries" and "cities". ;)


jpic:
Yeah but they didn't relate very well for the example :dancer: 
Thankssss

PR #598, opened 2016-02-18
Fix small typo in  documentation.

akatsoulas:


jpic:
Thankssss !! Just wondering if we shouldn't rather rename YourCountryCreateField to YourCountryCreateMultipleField above in this example because it inherits from autocomplete.CreateModelMultipleField, what do you think ?

akatsoulas:
@jpic makes total sense! I just amended the commit.

jpic:
:guitar:

PR #609, opened 2016-02-29
WIP Fix #608: V2-style on-the-fly object creation

### LINKED ISSUE #608: ValueError at [...] invalid literal for int() with base 10: 'test78'
Hi,
When trying to create a new item, but leaving another mandatory field empty, the following error appears: 
_ValueError at /admin/select2_one_to_one/testmodel/51/change/
invalid literal for int() with base 10: 'test78'_

This is easily reproductible on the demo site:
- Let the field "Name" empty
- Enter a non existing item in test field and select: Create "whatEverDoNotExit"
- Click on the Save button

It looks like this bug [was previously solved](https://github.com/yourlabs/django-autocomplete-light/issues/62), but it occurs again in current version.

Best

jpic:
It didn't work very well to create the object in the field's clean()
method. Another option was to do such thing in the widget's
value_from_datadict() method but there is no good way to know if a value
is meant to be an existing id or a new value anyway (ie. string ids).

This patch rewrites this feature in the same way it was implemented in
v2.

It breaks backward compatibility as it removes a few form field classes
in favor of a new argument in the view creation.

A new test app was added, select2_many_to_many_create, to test that
feature on multiple selects.

WIP: add tests and docs before merging.


src/dal/views.py
```diff
@@ -49,3 +59,27 @@ def get_queryset(self):
...
+    def post(self, request):
+        """Create an object given a text, for staff only."""
+        if not request.user.is_authenticated() and not request.user.is_staff:
+            # We should have security by default, but this should be somehow
+            # configurable. Perhaps check django.contrib.auth.permissions or
+            # reuse something that comes from django.contrib.auth ?
+            return http.HttpResponseForbidden()
+
+        if not self.create_field:
+            return http.HttpResponseForbidden()
```
kevin-brown: Wouldn't it make sense to raise [an `ImproperlyConfigured` exception](https://docs.djangoproject.com/en/1.9/ref/exceptions/#improperlyconfigured) here, with a note about setting the `create_field`?

src/dal_select2/static/autocomplete_light/select2.js
```diff
@@ -6,7 +6,8 @@
         $(this).find('option[value=""]').remove();
 
         $(this).select2({
-            tags: element.attr('data-tags'),
+            tags: element.attr('data-tags') ? true : false,
```
kevin-brown: Select2 _should_ [automatically parse `data-tags`](https://select2.github.io/options.html#data-attributes) and enable tagging.

src/dal_select2/static/autocomplete_light/select2.js
```diff
@@ -48,6 +49,43 @@
...
+                },
+                beforeSend: function(xhr, settings) {
+                    xhr.setRequestHeader("X-CSRFToken", document.csrftoken);
+                },
+                success: function(data, textStatus, jqXHR ) {
+                    var option = '<option value="' + data.id + '">' + data.text + '</option>'
+                    select.append(option);
+                    select.select2('close');
+
+                    if (select.attr('multiple') == 'multiple') {
```
kevin-brown: Instead of checking the attribute, it's recommended to check the underlying DOM property with `.prop('multiple')`. This also has the advantage of returning a boolean instead of (hopefully) `'multiple'`.

src/dal/views.py
jpic: If `create_field` is not set, then the autocomplete should not even post to this view. That said, ImproperlyConfigured is better, thanks !

src/dal_select2/static/autocomplete_light/select2.js
jpic: Good to know, thanks !

PR #611, opened 2016-03-01
Fix #610: Cancel out Django's style on lists

### LINKED ISSUE #610: Black square when renders selected items of a many to many autocomplete
Hi all,

I'm using version 3.0.4 of **dal** with **django** 1.8.8. I use it for some autocompletes in the admin and for many to many appears a strange black square:

dal-black-square

It only happens for the  many to many one. I have some normal autocompletes, ModelSelect2, and the black square does not appears. It is rendered on firefox and chrome browsers.

Somebody has experienced the same issue? Any idea of a possible solution?

Kind regards.

jpic:


jpic:
Upstream PR: https://github.com/select2/select2/pull/4226

If it's merged in a timely fashion, we should release the upstream fix instead of this patch.

PR #619, opened 2016-03-06
sbo_selenium to pytest+splinter

jpic:


jpic:
Can't believe it ain't faster to have a single browser window throughout all the test suite ...

Issue #635, opened 2016-03-11
v3 question: how to complete with list instead of queryset?


dmosberger:
Hopefully I'm missing something obvious, but it's not clear to me how to use a list of choices (rather than a queryset) for autocompletion.  For example, in v2, I could write:

```
class Foo_Autocomplete(al.AutocompleteListBase):
 choices = [ 'choice A', 'choice B', 'choice C' ]
 ...
```
How can I do the same with v3?

jpic:
It's not demonstrated yet, but you could use a ChoiceField instead of a ModelChoiceField and in the view use reuse the mixins you want.

Are you really using a list or are you using another backend (ie. redis) or something ?

dmosberger:
Yes, in the project I'd like to do that, we want to autocomplete, e.g., on the list of (electronic part) footprints.  Those footprints are generated for an EDA program (e.g., kicad), so it doesn't really make sense to represent them in a model.  Instead, we just construct a list of available footprints by walking a local directory containing them.

Any chance to see some sample code?  I'm not that familiar with the inner workings of dal3.

jpic:
Yeah perhaps, will you be able to contribute documentation if you have example code ?

dmosberger:
Sure, seems like a fair deal!

jpic:
PR uploaded !

Note that I did a couple of minor changes in the package, if you don't want to do them then you can import ViewMixin from dal.views, and put the Select2 class directly in your form (it contains no code).

jpic:
As mentioned in the PR, note that you don't even have to pass a url to the widget if you don't need a view ! It really depends, how much options do you want to propose ? If < 300 I'd say just go for :

```
class Select2(WidgetMixin, Select2WidgetMixin, forms.Select):
    """Select widget for regular choices and Select2."""


class TestForm(forms.ModelForm):
    test = forms.ChoiceField(
        choices=[
            ('windows', 'windows'),
            ('linux', 'linux'),
        ],
        required=False,
        widget=autocomplete.Select2(url='select2_list')
    )

    class Meta:
        model = TestModel
        fields = ('name', 'test')

```

Now feel free to go ahead and share some creativity to present that in nice documentation for the community with :heart: B)

dmosberger:
Cool, let me play with this to make sure I understand it.

dmosberger:
[...]

PR #637, opened 2016-03-11
Request #635: demo of list autocompletion

jpic:


jpic:
Also note that you don't even have to pass a url to the widget if you don't need to back it with a view !

jpic:
#745

PR #638, opened 2016-03-11
Check if submodules are cloned

jpic:
Blow up at the face of users trying to start dal_select2 without select2 :D

ie. #635

Should help a lot ^^

### DIFF
--- a/src/dal_select2/checks.py
+++ b/src/dal_select2/checks.py
@@ -1,0 +1,24 @@
+"""Checks for the dal_select2 module."""
+import os
+
+from django.core import checks
+
+
+def select2_submodule_check(app_configs, **kwargs):
+    """Return an error if select2 is missing."""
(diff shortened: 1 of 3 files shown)


src/dal_select2/checks.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+from django.core import checks
+
+
+def select2_submodule_check(app_configs, **kwargs):
+    errors = []
+    select2 = os.path.join(
+        os.path.abspath(os.path.dirname(__file__)),
+        'static/autocomplete_light/vendor/select2/dist/js/select2.min.js'
```
thijstriemstra: use os.path.join on all paths, this'll fail on windows.

jpic: Pull request welcome for windows support !

PR #644, opened 2016-03-15
Add autocomplete.Select2ListView class

dmosberger:
This class makes it easy to autocommplete based on a list of strings.

### DIFF
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -2,9 +2,10 @@
 
 import json
 
-from dal.views import BaseQuerySetView
+from dal.views import BaseQuerySetView, ViewMixin
 
 from django import http
+from django.views.generic.list import View, BaseListView
 
 
 class Select2ViewMixin(object):
@@ -44,3 +45,47 @@ class Select2ViewMixin(object):
 
 class Select2QuerySetView(Select2ViewMixin, BaseQuerySetView):
     """List options for a Select2 widget."""
+
+class Select2ListView(ViewMixin, View):
+    """Autocomplete from a list of items rather than a QuerySet.
+    get_list() should return a list of strings that may be autocopleted.
+    """
+
+    create_field = False
+    """Set create_field to True to allow creation of new items."""
+
+    def get_list(self):
+        """"Return the list strings from which to autocomplete."""
+        return []
+
+    def get(self, request, *args, **kwargs):
+        results = self.get_list()
+        create_option = []
+        if self.q:
+            results = [ x for x in results if self.q in x ]
+            if self.create_field:
+                create_option = [{
+                        'id': self.q,
+                        'text': 'Create "%s"' % self.q,
+                        'create_id': True
+                    }]
+        return http.HttpResponse(json.dumps({
+            'results': [dict(id=x, text=x) for x in results] + create_option
+        }))
+
+    def create(self, text):
+        """"Create a new list item for text.  None can be returned if the
+        item may not be created for some reason."""
+        return text
+
+    def post(self, request):
+        text = request.POST.get('text', None)
+        if text is None:
+            return http.HttpResponseBadRequest()
+        text = self.create(text)
+        if text is None:
+            return http.HttpResponseBadRequest()
+        return http.HttpResponse(json.dumps({
+            'id': text,
+            'text': text,
+        }))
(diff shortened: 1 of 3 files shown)


jpic:
Awesome, thank you so much !

Would it be possible to provide an example app too, ie. test_project/select2_list ?

Would you be able to add some tests too ?

Keep up the great work B)

dmosberger:
On Wed, Mar 16, 2016 at 10:58 AM, James Pic <notifications@github.com>
wrote:

> Awesome, thank you so much !
>
> Would it be possible to provide an example app too, ie.
> test_project/select2_list ?
>
> Would you be able to add some tests too ?
>
​I couldn't get the test_project to work because one of its dependencies
(maybe taggit) was depending on a feature that's no longer available in the
latest Django, so I gave up.​

​  --david​

-- 
Love Linux kernel hacking?  Know Python?  Need a job?  Talk to us at
jobs@egauge.net

jpic:
Could you paste the traceback ?

Did you try building a venv with tox ?

dmosberger:
No, I didn't know about tox.  That seems to work.
I'll be off for a week (Spring break) but will take a look after.

Thanks,

  --david

src/dal_select2/views.py
```diff
@@ -44,3 +45,47 @@ def render_to_response(self, context):
...
+                        'create_id': True
+                    }]
+        return http.HttpResponse(json.dumps({
+            'results': [dict(id=x, text=x) for x in results] + create_option
+        }))
+
+    def create(self, text):
+        """"Create a new list item for text.  None can be returned if the
+        item may not be created for some reason."""
+        return text
```
jpic: Shouldn't we raise NotImplemented or something like that here ?

dmosberger: On Sun, Mar 27, 2016 at 10:03 PM, James Pic <notifications@github.com>
wrote:

> In src/dal_select2/views.py
> <https://github.com/yourlabs/django-autocomplete-light/pull/644#discussion_r57543419>
> :
>
> > +        if self.q:
> > +            results = [ x for x in results if self.q in x ]
> > +            if self.create_field:
> > +                create_option = [{
> > +                        'id': self.q,
> > +                        'text': 'Create "%s"' % self.q,
> > +                        'create_id': True
> > +                    }]
> > +        return http.HttpResponse(json.dumps({
> > +            'results': [dict(id=x, text=x) for x in results] + create_option
> > +        }))
> > +
> > +    def create(self, text):
> > +        """"Create a new list item for text.  None can be returned if the
> > +        item may not be created for some reason."""
> > +        return text
>
> Shouldn't we raise NotImplemented or something like that here ?
>
I don't feel very strongly about it.  I thought that if you set
create_field to True, you'd expect that any value would be accepted for
creation by default.  What's your thought process?

  --david​
-- 
Love Linux kernel hacking?  Know Python?  We are hiring!  Contact us at
jobs@egauge.net

jpic: I think the verb "create" implies some new object is persisted somewhere. Perhaps you could describe why that works for as-is ?

dmosberger: On Tue, Apr 5, 2016 at 2:50 AM, James Pic <notifications@github.com> wrote:

> In src/dal_select2/views.py
> <https://github.com/yourlabs/django-autocomplete-light/pull/644#discussion_r58504780>
> :
>
> > +        if self.q:
> > +            results = [ x for x in results if self.q in x ]
> > +            if self.create_field:
> > +                create_option = [{
> > +                        'id': self.q,
> > +                        'text': 'Create "%s"' % self.q,
> > +                        'create_id': True
> > +                    }]
> > +        return http.HttpResponse(json.dumps({
> > +            'results': [dict(id=x, text=x) for x in results] + create_option
> > +        }))
> > +
> > +    def create(self, text):
> > +        """"Create a new list item for text.  None can be returned if the
> > +        item may not be created for some reason."""
> > +        return text
>
> I think the verb "create" implies some new object is persisted somewhere.
> Perhaps you could describe why that works for as-is ?
>
​This mirrors the model-based version: create() should create a new
"object" for text, except in the simplest case, the text itself is the new
object and no other action needs to be taken.

For example, in EPIC, we autocomplete based on the part manufacturers name,
but we want to let the user enter arbitrary new manufacturer names for new
parts, so the above default does what we need just fine.

  --david

-- 
Love Linux kernel hacking?  Know Python?  We are hiring!  Contact us at
jobs@egauge.net

jpic:
Closing in favor of #702, could you check that it works for you please ? I've decided to take a bit of technical debt and would like this to be in the next release even without tests. Thanks !!

PR #649, opened 2016-03-16
No need to use -e

andybak:
No need to use -e if all you need is to install Git master. It's an extra bit of complexity and can cause problems that a normal pip install wouldn't.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -8,9 +8,9 @@ Install version 3 with :pip:ref:`pip install`::
 
     pip install django-autocomplete-light
 
-Or, install the dev version with git::
+Or, install the dev version from Github::
 
-    pip install -e git+https://github.com/yourlabs/django-autocomplete-light.git#egg=django-autocomplete-light
+    pip install https://github.com/yourlabs/django-autocomplete-light/archive/master.zip
 
 Then, let Django find static file we need by adding to
 :django:setting:`INSTALLED_APPS`, **before** ``django.contrib.admin``::


docs/install.rst
```diff
@@ -8,9 +8,9 @@ Install version 3 with :pip:ref:`pip install`::
 
     pip install django-autocomplete-light
 
-Or, install the dev version with git::
+Or, install the dev version from Github::
 
-    pip install -e git+https://github.com/yourlabs/django-autocomplete-light.git#egg=django-autocomplete-light
+    pip install https://github.com/yourlabs/django-autocomplete-light/archive/master.zip
```
jpic: If you install the dev version this way then you can't develop in the package to contribute without re-installing every time you change a file.

Perhaps we could mention both ways of installing ?

andybak: Yep. No harm having both. I just wanted to correct the implication that -e was the only way to install from Github (I thought this to be the case for a while myself!)

Do you want another PR or shall I leave it to you?

jpic: On Wed, Mar 16, 2016 at 11:37 AM, Andy Baker <notifications@github.com>
wrote:

> Yep. No harm having both. I just wanted to correct the implication that -e
> was the only way to install from Github (I thought this to be the case for
> a while myself!)
>
> Do you want another PR or shall I leave it to you?
>
> It would be great if you updated the pr !

Thanks !

jpic:
I don't think this works, because it won't install submodules, and we need the select2 submodule to work.

andybak:
I'm not sure I fully understand.

Consider three ways to install:
1. pip install django-autocomplete-light
2. pip install https://github.com/yourlabs/django-autocomplete-light/archive/master.zip
3. pip install -e git+https://github.com/yourlabs/django-autocomplete-light.git#egg=django-autocomplete-light

Aren't 2+3 more similar to each other than they are to 3? What is it that works with 1 and 3 that doesn't work with 2?

jpic:
The problem with 2. is that select2 won't be installed:

```
[newenv] 18/05 2016 00:42:45 jpic@lue /tmp/newenv/lib/python2.7/site-packages/dal_select2/static 
$ find .
.
./autocomplete_light
./autocomplete_light/select2.css
./autocomplete_light/select2.js
```

Vs.:

```
$ find src/dal_select2/static/autocomplete_light/
src/dal_select2/static/autocomplete_light/
src/dal_select2/static/autocomplete_light/vendor
src/dal_select2/static/autocomplete_light/vendor/select2
src/dal_select2/static/autocomplete_light/vendor/select2/select2.jquery.json
src/dal_select2/static/autocomplete_light/vendor/select2/.travis.yml
src/dal_select2/static/autocomplete_light/vendor/select2/README.md
[...]
```

So the Widget.Media's links to `autocomplete_light/vendor/select2/....` are broken.

jpic:
It's a common problem with github tarballs + submodules, you'll run into the same if you try to install ansible from github's tarball ;)

jpic:
That said, you still don't need -e afaik, but installing from the tarball won't get you the git submodule to select2 !

PR #656, opened 2016-03-23
Use http.JsonResponse() to return json

guettli:


jpic:
Really cool, thanks !

PR #665, opened 2016-03-31
Quit testing taggit on dj110

jpic:
See https://github.com/alex/django-taggit/issues/384


jpic:
The commit that required this was reverted in django.

jpic:
The revert was not complete in django.

jpic:
Fixed in django + we pass on dj110

PR #679, opened 2016-04-27
allow create message translation/customization

maisim:
Hi,
Thanks for your work on dal.
This little change allow to translate the message displayed by dal to create new item.

### DIFF
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -5,6 +5,7 @@ import json
 from dal.views import BaseQuerySetView
 
 from django import http
+from django.utils.translation import ugettext as _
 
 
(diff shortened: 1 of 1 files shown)


jpic:
Thanks a lot for your feedback @maisim, would it be possible for you to provide a translation file as well ?

maisim:
Yes, for the moment my way to use it is to symlink dal directory in my project to get "manage.py makemessages" able to find strings to translate and add them to my project translations files. But it should be nice to get translation for everywhere it's needed in dal. I will look for it asap.

jpic:
Thanks that would be awesome ! Where are you from Simon ?

Issue #677, opened 2016-04-18
Forward data not sent when creating object


CreatixEA:
When using the forward option, and when there is no result and when we can create an object : the forwarded data are not sent to the creation function

Inside `create_object(self, text)` the `self.forwarded` is always `None`. IMO it should be filled with the forwarded data. It's a JS trick I think in that [dict](https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal_select2/static/autocomplete_light/select2.js#L82)

Hope it will be implemented and thank you for that wonderfull tool.

Sincerely,

E.

jpic:
Could you make an example app for this please ?

Thanks !

toudi:
I have figured out how to do this, I'll do a PR in a moment.

toudi:
could we please close this one, @jpic ?

jpic:
Do you have permission to close it now ?

toudi:
yep, it works ;) thanks!

jpic:
:guitar:

PR #682, opened 2016-05-05
fix for #677

toudi:
extend 'forward' attribute (linked data feature) to creating objects as well


jpic:
I'd say you're making history right now

jpic:
What can I say, you're tearing it up.

PR #690, opened 2016-05-17
Fix #678: clean autocomplete on forward field change

### LINKED ISSUE #678: clean "autocomplete field" if "Forward field" has changed
Hello, I have the following use case (is a variation of http://django-autocomplete-light.readthedocs.org/en/master/tutorial.html#filtering-results-based-on-the-value-of-other-fields-in-the-form )

if a "Forward field" has changed, the "autocomplete field" should be cleaned. 
this would avoid the user to select options incompatible with the currently selected filter.

currently I'm using this little code snippet that I insert on https://github.com/yourlabs/django-autocomplete-light/blob/3.1.5/src/dal_select2/static/autocomplete_light/select2.js#L7

```javascript
        (function (autocomplete_element) {
            var forward = $(autocomplete_element).attr('data-autocomplete-light-forward');
            if (forward !== undefined) {
                forward = forward.split(',');

                var parts = $(autocomplete_element).attr('name').split('-');
                var prefix = '';

                for (var i in parts) {
                    var test_prefix = parts.slice(0, i).join('-');
                    if (!test_prefix.length) continue;
                    test_prefix += '-';

                    if ($(':input[name=' + test_prefix + forward[0] + ']').length) {
                        var prefix = test_prefix;
                    }
                }

                for (var key in forward) {
                    var name = prefix + forward[key];
                    var element_selector_str = '[name=' + name + ']';
                    // attach change event on "forward fields" to clean
                    // autocomplete_element if "forward field" has changed
                    $('[name=' + name + ']').on('change', function () {
                        // this clean current selection on autocomplete field
                        $(autocomplete_element).empty().trigger('change')
                    });
                }
            }

        })(this);

```

The question is:

1 - This should be the default behavior?
2 - An attribute like as `data-autocomplete-light-clean-if-forward-changed` should be provided to enable/disable this behavior?

What are your thoughts about this?

ps. I did not like this code snippet because it is essentially https://github.com/yourlabs/django-autocomplete-light/blob/3.1.5/src/dal_select2/static/autocomplete_light/select2.js#L26-L53 with few modifications (almost duplicate code).. but it was the best I could do with my current knowledge of javascript ...

I tries to send a pull-request if you want

jpic:


Fabitosoft:
Hi again @jpic 

I couldn't figure out how to use the autoclean feature when forward change. Do I have to put something in the form like option? It should work by its own automatically?

Thanks

jpic:
It's in JS: http://django-autocomplete-light.readthedocs.io/en/master/tutorial.html?highlight=clearing#clearing-autocomplete-on-forward-field-change

s0th1nk:
If you use bootstrap 4 with  jQuery’s slim version and have
`<script type="text/javascript" src="{% static 'admin/js/vendor/jquery/jquery.js' %}"></script>`
in script block
this script
```
$(document).ready(function() {
    $(':input[name$=owner]').on('change', function() {
        var prefix = $(this).getFormPrefix();
        $(':input[name=' + prefix + 'test]').val(null).trigger('change');
    });
});
```
will not work

To fix the issue
remove 
`<script type="text/javascript" src="{% static 'admin/js/vendor/jquery/jquery.js' %}"></script>`
and use full version of bootstrap 4 jquery ( jquery-3.4.1.min.js )

saranmaikalluri:
Not able to clear autocomplete field with the above method.. am not using bootstrap4; any form level option available or custom js to clear the field on forward field change?

PR #691, opened 2016-05-17
Fix #670: describe JS loading

### LINKED ISSUE #670: UseCase: I want select2, but not autocomplete
Hi,

does django-autocomplete-light provide a django-widget, which does apply select2-JS-magic to the HTML-select, but does not use an autocomplete?.

My use case: I want a better ManyToMany-Widget, but it should be a drop-in replacement. There are only very few (less then 12) items in the list.

I can live with ctrl-click, but most users don't get it :-)

jpic:


jpic:
Thanks for the review @guettli !

PR #696, opened 2016-05-30
Select2 without ajax

jpic:
I need to add tests before merge and release but FYI I'm using this branch in production.


jpic:
Added in 5c4ba9a

PR #705, opened 2016-07-10
improve compatibility for django 1.9

9nix00:
django 1.9 changed  jquery.init.js file.

you can see [this](https://github.com/django/django/blob/ec4f219ecb7a5e43d0353633fac4dac42d0ee492/django/contrib/admin/static/admin/js/jquery.init.js)

the old code in release package is 
```
var django = django || {};
django.jQuery = jQuery

var yl = yl || {};
yl.jQuery = django.jQuery

// this file also prevents django.jQuery.noConflict
// there's got to be a better way ...
```
use current code will cause jQuery is undefined. it looks we have a scope error.

this PR can make our code stronger.

and I am asking why django team change this init file. if they won't plan rollback to old style. 
we have to add our jquery file. so sad.

### DIFF
--- a/src/dal/static/autocomplete_light/autocomplete.init.js
+++ b/src/dal/static/autocomplete_light/autocomplete.init.js
@@ -60,7 +60,7 @@ element was cloned with data - which should be the case.
         if (document.cookie && document.cookie != '') {
             var cookies = document.cookie.split(';');
             for (var i = 0; i < cookies.length; i++) {
-                var cookie = yl.jQuery.trim(cookies[i]);
+                var cookie = $.trim(cookies[i]);
                 // Does this cookie string begin with the name we want?
                 if (cookie.substring(0, name.length + 1) == (name + '=')) {
                     cookieValue = decodeURIComponent(cookie.substring(name.length + 1));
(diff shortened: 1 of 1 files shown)


src/dal/static/autocomplete_light/autocomplete.init.js
```diff
@@ -60,7 +60,7 @@ element was cloned with data - which should be the case.
         if (document.cookie && document.cookie != '') {
             var cookies = document.cookie.split(';');
             for (var i = 0; i < cookies.length; i++) {
-                var cookie = jQuery.trim(cookies[i]);
+                var cookie = yl.jQuery.trim(cookies[i]);
```
jpic: Thanks a LOT !

Are you sure we shouldn't have $ here instead ?

9nix00: I planed use $, but I found django admin javascript use full name. so I follow them style.

jpic: I think using $ would make the code more portable here, care to change it ?

9nix00: hi,@jpic. upgrade.

but strictly speaking for now,this PR's real target is improve this file code style. the old topic is not right.

because finally, I found this error is cause by myself.I am so sorry.
I used django-autocomple-light  in others project with django 1.8,they works well . 
I create a new project with django1.9. but I load `dal` and `dal_select2` after `django.contrib.admin`
so error caused. and I found this file use jQuery not $ or yl.jQuery.

then I compare this project with others project in browser. I found jquery.init.js is different.
 
the truth is I compare jquery.init.js file from dal package  with  django.contrib.admin package's  jquery.init.js file. so, I trapped 😫...

jpic:
Thanks !

PR #706, opened 2016-07-11
setup.py: use lowercase and no space with `extras_require`

blueyed:
Ref: https://github.com/nvie/pip-tools/pull/354


jpic:
Thanks !

PR #714, opened 2016-07-22
Update select2.js

meesterguyman:
Currently, inline forms are unable to access forward fields from the main form. By adding this fallback, inlines will prefer forward fields within their own forms but will be able to access forward fields outside of themselves (not having the inline form prefix) when no such fields are found within the inline.


jpic:
Nice one, thanks !

PR #718, opened 2016-07-26
Remove code block related to select2 upstream issue

FuzzAU:
The issue discussed the block refers to https://github.com/select2/select2/pull/4249 which has since been closed and resolved with another issue.

This block causes issues in IE11/Edge. When the change was triggered and the selection__clear element was added, this block of code would remove it again.


FuzzAU:
It seems I broke some things, I'll go back and investigate.

Is there a preference for closing the issue and launching a new one, or just updating this PR?

jpic:
Go ahead and add commits to this PR, we'll rebase before merging.

jpic:
OK, I think it's safe to remove this code block as long as we upgrade the select2 submodule. Could you please do that ?

Thanks !!

FuzzAU:
Hmm updated to the commit for the latest select2, but same test fails. I don't suppose you're familiar with this test @jpic ?

jpic:
Yep, this breaks test_can_change_selected_option which tests updating an option label via the edit button which opens a popup in django admin. Basically, the piece of code you're removing is in charge of updating the select2 label when the selected option's inner html change, which is the result of updating the model in a popup. The good news is that we can reproduce the issue locally, so it's not related to the environment on travis itself.

FuzzAU:
Yes the first thing I did was run the tests locally too. I guess the suggestion here is that the commit to select2 referenced in the comments are not sufficient to the remove the code fragment.

The issue that bought me here was that the behaviour of calling text($(this).text()); on IE actually removes selection__clear immediately after it gets added.

jpic:
FTR we have another issue upstream for that: https://github.com/select2/select2/issues/4733

jpic:
Tests pass without the hack after select2 update ;)

PR #721, opened 2016-07-27
note about placement of DAL before grappelli

chubz:


docs/install.rst
```diff
@@ -23,10 +23,12 @@ This is to override the ``jquery.init.js`` script provided by the admin, which
 sets up jQuery with ``noConflict``, making jQuery available in
 ``django.jQuery`` only and not ``$``.
 
-Note: If you are trying to install from git, please make sure you are not using
+Notes:
+-If you are trying to install from git, please make sure you are not using
 **zip/archive** url of the repo ``django-autocomplete-light`` since it will not
 contain required submodules automatically. Otherwise these submodules will then
 need to be updated separately using ``git submodule update --init``.
+-If you are using django-autocomplete-light you need to also place ``dal``,``dal_select2`` ... above/before grappelli
```
jpic: Could you wordwrap please ?

With vim you can use gqq, with other text editor dunno ^^

jpic: Also, I think this should not be here, but rather replace:

> Then, let Django find static file we need by adding to :django:setting:`INSTALLED_APPS`, **before** django.contrib.admin:

With


> Then, let Django find static file we need by adding to :django:setting:`INSTALLED_APPS`, **before** django.contrib.admin **and** grappelli if present:

chubz: Sounds fine! Do you think that maybe then grappelli should be added above 'django.contrib.admin', in the installed app example also?
maybe just be commented out like this:
Then, let Django find static file we need by adding to
:django:setting:`INSTALLED_APPS`, **before** ``django.contrib.admin`` and
``grappelli`` if present::

    'dal',
    'dal_select2',
    # 'grappelli',
    'django.contrib.admin',

jpic: yes :guitar:

chubz: Added the second commit ... should be as we agreed now! Thanks for the comment!

jpic: Should it rather be "If you are using grappeli" ? Also, could you wrap words please ?

Thanks a lot !!

:guitar:

jpic:
:guitar:

Issue #729, opened 2016-08-30
runserver on test_project raises SystemCheckError that prevents launching


thecardcheat:
When running

`python test_project/manage.py runserver 192.168.1.177:8000`

I get

```
django.core.management.base.SystemCheckError: System check identified some issues:

ERRORS:
select2_generic_foreign_key.TestModel.content_type: (fields.E304) Reverse accessor for 'TestModel.content_type' clashes with reverse accessor for 'TestModel.test'.
        HINT: Add or change a related_name argument to the definition for 'TestModel.content_type' or 'TestModel.test'.
```

My Django is 1.10.

Could use confirmation of reproduction since it is not a previously reported issue.

Will submit a simple PR to resolve this momentarily.

PR #730, opened 2016-08-30
Resolve SystemCheckError fields.E304

thecardcheat:
#729


jpic:
Way to go, thanks !

:guitar:

PR #733, opened 2016-09-01
Advanced forward features

gagarski:
This commit adds arrow notation for client-side forwarded fields renaming.
Let's suppose you have a form with fields `owner` and `test`. `test` autocomplete forwards field `owner` and server side expects `owner` as a key for the forwarded value.
Then you want to create another form that uses the same autocomplete for `test`, but for some reason you want to forward a field `master` as you would do with `owner`. 
Now you can declare forward in the second form as `master->owner` which means "Take a value from field `master` and forward it as `owner`". An autocomplete view used by the first form now can be reused in the second form.
Of course, you still can declare forwarded value without arrow notation: `owner` in a forward declaration is equivalent to `owner->owner`.


jpic:
Hi !

Thanks for your contribution !

I find the arrow notation a bit "funky", apart from that this feature
addresses an issue that makes sense. That said, we'd need to cover it with
tests and documentation before we can release it, that also means more
maintainance in the future. Perhaps it would be a good idea to explore some
alternative solutions to this issue before going forward with this.

For example, I'm wondering if we could perhaps just forward arbitrary data
in addition to form fields. That kind of feature could have solved a few
use case that have been brought up already, and that were solved by adding
a hidden form field purely for forwarding. For example, for your use case,
it would look like:

    class PersonForm(forms.ModelForm):
        class Meta:
            widgets = {
                'field': autocomplete.ModelSelect2(
                    url='country-autocomplete',
                    forward=['continent'],
                    data={'type': 'person'},
                )
            }

    class CompanyForm(forms.ModelForm):
        class Meta:
            widgets = {
                'field': autocomplete.ModelSelect2(
                    url='country-autocomplete',
                    forward=['continent'],
                    data={'type': 'company'},
                )
            }

Then, in the autocomplete view, you could switch with:

    continent = self.forwarded.get('continent', None)
    form_type = self.data.get('type', None)

I'm also wondering if you wouldn't be better off with 2 autocomplete urls,
one for each form, to have one autocomplete class per form, leveraging
simple python inheritance to keep the code separated rather than having one
class with switches in it.

gagarski:
> I'm also wondering if you wouldn't be better off with 2 autocomplete urls

It's not very nice to have two different views those perform exactly one task. Let's say, the task is "Give me please countries on a given continent". 
To do this you can forward a `continent` field to a view `CountryAutocomplete`. It works until your form field is named `continent`. Let's consider another example:
```python
class ShippingForm(forms.Form):
    src_continent = forms.ModelChoiceField(
        queryset=Continent.objects.all(),
        widget=autocomplete.ModelSelect2(url='continent_autocomplete'))
    src_country = forms.ModelChoiceField(
        queryset=Country.objects.all(), 
        widget=autocomplete.ModelSelect2(
            url='country_autocomplete',
            forward=('src_continent->continent',)))
    src_city = forms.ModelChoiceField(
        queryset=City.objects.all(),
        widget=autocomplete.ModelSelect2(
            url='city_autocomplete',
            forward=('src_country->country',)))
    # More source fields...
    dst_continent = forms.ModelChoiceField(
        queryset=Continent.objects.all(),
        widget=autocomplete.ModelSelect2(url='continent_autocomplete'))
    dst_country = forms.ModelChoiceField(
        queryset=Country.objects.all(),
        widget=autocomplete.ModelSelect2(
            url='country_autocomplete',
            forward=('dst_continent->continent',)))
    dst_city = forms.ModelChoiceField(
        queryset=City.objects.all(),
        widget=autocomplete.ModelSelect2(
            url='city_autocomplete',
            forward=('dst_country->country',)))
    # More destination fields...
```

This allows us to have three autocomplete views: 
 - `ContinentAutocompleteView` without forwarded values;
 - `CountryAutocompleteView` with `continent` as a forwarded value;
 - `CityAutocompleteView` with `country` as a forwarded value.

Alternative approaches are:
 - Have two autocomplete views for each Country and City: one that expects `src_` value and one that expects `dst_value`. Not so good.
 - Have one autocomplete view for each Country and City, that expects `src_` and `dst_` forwarded value. Better, but these views should be aware of all forms using them.

jpic:
Interesting ! I didn't understand your use case in that way. Anyway, it's
true that it would add more boilerplate code but perhaps have more explicit
separation of concern:

```
CountryAutocomplete:
    country_field = 'country'

    def get_country(self):
        return self.forwarded.get(self.country_field, None)

    def get_queryset(self):
        country = self.get_country()
        ...


SrcCountryAutocomplete(CountryAutocomplete):
    country_field = 'src_country'


DstCountryAutocomplete(CountryAutocomplete):
    country_field = 'dst_country'
```

Did you consider it that way ? I can see how that will throw in a handful
more lines of code in your project at the end of the day, but the upside
for you are:

- no need to redact additional documentation in DAL,
- no need to code additional tests in DAL,
- no need to commit to maintain this feature "until the end of your life"
in DAL, or something like that :)
- a lot less work for you, potentially, assuming I understand your issue
correctly, which might not be the case still, keep in mind i'm at the other
end of the world xD

If, having considered all that, you are still convinced to contribute this
to DAL, then read on ;)

For the "funky" notation, we could replace it with a Forward class that
could take both args and kwargs, ie.

    ['a', 'b->c']

Would instead be

    Forward('a', b='c')

But I'm not really convinced by that approach either, because that seems a
bit like overkill.

In any case, if we were going for something like this patch, did you
consider using `:` instead of `->` ? That would be like the git syntax, ie.
git push origin mymaster:master which pushes local branch mymaster on the
origin's master branch, and also ressembles yaml and json and python dict
key:value syntax.

Perhaps @blueyed would be interrested in sharing an opinion on this too.

jpic:
Ok, scratch the first part of my above comment, I completely understand the
necessity of this now !

gagarski:
>- a lot less work for you, potentially, assuming I understand your issue
correctly, which might not be the case still, keep in mind i'm at the other
end of the world xD

Actually, my example with Continent/Country/City is a bit artificial. I am using auto-generated forms for models (something like ModelForm but for filtering) and field names are sometimes looks ugly (like `city__country__continent` or `parcel__dst_city__country__continent` etc.). Supporting all these names on server side looks like hell for me:)

> In any case, if we were going for something like this patch, did you
consider using `:` instead of `->` ?

The choice of `->` was completely arbitrary and supposed to denote semantics of `to` preposition. But now I understand, that it would be quite painful if someone sometimes will try to extract form prefix from forwarded keys. `=>`, `>>` or `|>` probably would be a better choice, also I am okay with `:`. 

> For the "funky" notation, we could replace it with a Forward class that
> could take both args and kwargs, ie.
> 
>     ['a', 'b->c']
> 
> Would instead be
> 
>     Forward('a', b='c')
> 
> But I'm not really convinced by that approach either, because that seems a
> bit like overkill.

Also a list should be accepted as a forward argument if you do not want to break backwards compatibility.

 > - no need to code additional tests in DAL,

Functional tests from `linked_data` app in the test project seem to be good for "funky"-notation. Is it OK create a clone of this app but with funky notation?

jpic:
Absolutely, lets find a more relevant name than funky  though, I'm sorry I
came up with that LOL.

Go ahead with the app copy ! Perhaps rename_forward would be a relevant
name.

krikunts:
> For example, I'm wondering if we could perhaps just forward arbitrary data in addition to form fields

We had data attribute in version 2 (AFAIK) and this feature was really handy in many situations ([SO question related to this](http://stackoverflow.com/questions/25409293/django-autocomplete-light-add-parameter-to-query-url)). Since moving to v 3 I wonder why did you drop this feature.

gagarski:
Seems like I am done with tests and docs.

So, should we change `->` to something else?

I see the following alternatives:
- Other arrows: `src=>dst`, `src|>dst` (ugly one, as for me), `src>>dst`, `src>dst`. Pretty much the same as `->`. The direction of the rename is shown with an arrow.
- Assignment notation: `dst=src`. Notice swapped `src` and `dst`. The direction of the assignment should be obvious for user. 
- Reverse arrow notation (combining assignment and arrow notation): `dst<-src`, `dst<=src` etc.
- Colon notation: `src:dst` or `dst:src`. For me the direction of colon notation is a bit unclear.

jpic:
I suggest that we release this as is for now and then wait until we have a contribution for the class based system that would allow passing arbitrary data. 

@krikunts The problem with v2 is that it grew too fast with too much involvement from me ending in a situation where I only could maintain it. "If you want to go fast go alone, if you want to go far go together" ;)

So, I suggest that we release this as is for now, because it's better than nothing, then perhaps someone will start a contribution for the class based system that would also allow more features such as arbitrary values and reuse, in which case we'll deprecate this notation at some point. Unless @gagarski objects and prefers to collaborate on the class version now.

docs/tutorial.rst
```diff
@@ -407,6 +407,225 @@ filter as such in the view:
...
+            queryset=Country.objects.all(),
+            widget=autocomplete.ModelSelect2(
+                url='country-autocomplete',
+                forward=('dst_continent',)))
+        dst_city = forms.ModelChoiceField(
+            queryset=City.objects.all(),
+            widget=autocomplete.ModelSelect2(
+                url='city-autocomplete',
+                forward=('dst_country',)))
+        # More destination fields...
```
jpic: Wouldn't a minimalist example be better for the reader here ?

src/dal_select2/static/autocomplete_light/select2.js
```diff
@@ -8,13 +8,24 @@
...
+                var src_name, dst_name;
+                if (f.length === 0) {
+                    continue;
+                } else if (f.length === 1) {
+                    src_name = f[0];
+                    dst_name = f[0];
+                } else {
+                    src_name = f.splice(0,1);
+                    dst_name = f.join("->");
+                }
```
jpic: That works but when we implement a more generic data system we should use a json data structure or something instead of this, and have it encapsulated in a function in autocomplete.js

docs/tutorial.rst
gagarski: Like limiting fields to country/city? Or to only one continent/country/city field set? Or both?

src/dal_select2/static/autocomplete_light/select2.js
gagarski: We can use JSON but we probably should not store it in an HTML attribute because it involves ugly escaping/unescaping code.
A possible way to handle this is something like that:
```html
<select><!--with some select2 and dal attrs-->
  <!-- Here we declare three forwards:
    - forward field_name value without renaming
    - forward field_name2 value as forward_name
    - forward constant 42 as const_forward_name
  -->
  <script type="text/forwardconfig">
    [
      {
      	"type": "field",
      	"src": "field_name"
      },
      {
      	"type": "field",
      	"src": "field_name2",
      	"dst": "forward_name"
      },
      {
        "type": "const",
        "value": 42
        "dst": "const_forward_name"
      }
    ]
  </script>
  <!-- Select options -->
</select>
```

gagarski:
> Unless @gagarski objects and prefers to collaborate on the class version now.

I think, I can implement it. Now I see it this way:
```python
autocomplete.ModelSelect2(
    url='dal-url',
    forward=(
    	'field1'  # Old-fashioned way
    	Field('field2'),  # Same as above but with new syntax
    	Field('field3', dst="field30"),  # With rename. dst parameter is optional
    	Const('hello', dst="const_field")  # Constant forwarding. dst parameter is required.
    	)
    )
```
Is it OK?

src/dal_select2/static/autocomplete_light/select2.js
gagarski: > and have it encapsulated in a function in autocomplete.js

I think the whole add_forwards() function should be outside `select2.js` since it has no select2-specific functionality.

docs/tutorial.rst
jpic: Would it work to show just one field and rename it ? I don't think we have
to prove the use case in the tutorial, just show the feature in the most
minimalist way.

src/dal_select2/static/autocomplete_light/select2.js
jpic: That looks pretty cool indeed, we could also have a js callback type.

I know that storing json in html data attributes doesn't work very well
indeed. However, we shouldn't have that script tag inside the select field
because we want to support JAL which uses an input tag instead of a select.

Do you think it will work to have the script tag after the select tag ?

gagarski: > Do you think it will work to have the script tag after the select tag ?

Yes, I have already implemented it in next changesets.

> we could also have a js callback type.
Actually, the content of script tag with custom `type` value is an arbitrary text which we parse as JSON when we are adding forward values (see new changes). If we need callbacks we can do some dirty tricks with `eval()` function, but it does not seem a good idea for me now.

jpic:
Nothing worth delaying this to say, good work !

:guitar:

PR #734, opened 2016-09-04
Move create_option functionality into it's own method

andybak:
Main motivation was wanting to override this behaviour without touching any other parts of render_to_response but I think this is also a clearer way to organise the code.

Actually - in most cases all I really want to change is the wording of the create text. How would you feel about a get_create_label(self, queryset) method?

(example use case - I if someone enters a string with an @ in it I create a new contact with an email address, otherwise I assume it's a name - I'd like to make this behaviour explicit via the create label)

### DIFF
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -20,24 +20,30 @@ class Select2ViewMixin(object):
             } for result in context['object_list']
         ]
 
-    def render_to_response(self, context):
-        """Return a JSON response in Select2 format."""
+    def get_create_option(self, context, queryset):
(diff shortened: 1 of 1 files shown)


andybak:
That will teach me to rush. I should set up tox locally and fix the pep8 issue before making any more commits. Before I do so are you in favour of this patch in principle?

jpic:
Seems pretty same, that works for me, thanks for your contribution !

jpic:
Merged in #744, welcome in the team, keep up the great work B)

:guitar:

PR #737, opened 2016-09-09
TaggitSelect2: insure there's a comma when there's only one tag, or tag "Multi word" would end up as "Multi" and "word"

fruitschen:
When we enter only one multiple words tag and it will become multiple tags. Adding a comma in this case is harmless and will fix the problem. 
For example, Taggit will save "Multi Word" as "Multi" and "Word" 
"Multi Word, " will be treated correctly as "Multi Word"


andybak:
Just to clarify - the behaviour that is previewed before the tag is saved (i.e. the coloured block around the tag) indicates that the behaviour after save is going to be 'create a single tag'.

However after the tag is saved you end up with two tags. So an alternate fix might be to alter the way tags are displayed before saving?

jpic:
I like this, could you implement a simple test case ?

Thanks !!

:guitar:

fruitschen:
I've added a test case.

jpic:
Nice one @fruitschen !

@andybak 
> Just to clarify - the behaviour that is previewed before the tag is saved (i.e. the coloured block around the tag) indicates that the behaviour after save is going to be 'create a single tag'.

Currently [it seems broken](http://dal-yourlabs.rhcloud.com/admin/select2_taggit/testmodel/2/change/): it shows one tag to be created for `foo bar` but after saving we get two tags.

> However after the tag is saved you end up with two tags. So an alternate fix might be to alter the way tags are displayed before saving?

Doesn't this patch fixes exactly the server-side behaviour to match the client-side behaviour ?

If I'm not missing anything @andybak shares about then I'm all for releasing this !

andybak:
I was actually agreeing with @fruitschen  but pointing out that there are two fixes - fix the preview or fix what gets saved. I'm happy with this.

PR #741, opened 2016-09-19
Fix initial value not set when selected choices aren't strings

thauk-copperleaf:
Stringify values in selected_choices to match stringification of keys in filter_choices_to_render, so non-string types can be used as selected_choices (e.g. UUIDs)


jpic:
Nice one, thanks !

PR #743, opened 2016-09-21
Fix placeholder not working when URL is not given to autocomplete widget

thauk-copperleaf:
Correctly show an empty option tag when a placeholder is specified, which is what the select2 library expects by design, which fixes the placeholder not used when the widget is supplied with a list of choices but no external datasource URL.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -183,6 +183,9 @@ Passing options to select2
         },
     )
 
+.. note:: Setting a placeholder will result in generation of an an empty
+          ``option`` tag, which select2 requires.
+
 Using autocompletes in the admin
 ================================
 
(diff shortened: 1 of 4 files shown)


jpic:
Ok, but we need tests and documentation (a section in the tutorial, the smallest possible please).

thauk-copperleaf:
Can you give some guidance as to where you want the test to live?

I'm also not 100% certain how to set up locally to run the code coverage. Here's what I've done, assuming test_project is the correct place to run the tests:
- Created a virtual environment named venv
- Ran `pip install -r requirements.txt`
- Ran `pip install -e ../`
- Exported `DJANGO_SETTINGS_MODULE=settings.base` to my environment
- Ran `py.test --cov --liveserver 127.0.0.1:9999`

I don't think that's enough because I see a lot of things failing, but really all I need to do is just run my test (wherever it goes) to see if it improves code coverage.

thauk-copperleaf:
My test looks good, I'm just not sure if I put it in the right file.

jpic:
:guitar:

Issue #731, opened 2016-08-31
How to set language for select2?


sandino:
Hello, I use django-autocomplete-light==3.1.8 and it works fine. I also would like to make use of i18n features for Select2 widget but didn't find init script to set the language.

Could you please tell me how to set the language up?

gagarski:
A quick and dirty way to do it is to create your own widget based on `ModelSelect2`:
```python
class Select2RuWidgetMixin(Select2WidgetMixin):
    class Media:
        """Automatically include static files for the admin."""

        css = {
            'all': (
                'autocomplete_light/vendor/select2/dist/css/select2.css',
                'autocomplete_light/select2.css',
            )
        }
        js = (
            'autocomplete_light/jquery.init.js',
            'autocomplete_light/autocomplete.init.js',
            'autocomplete_light/vendor/select2/dist/js/select2.full.js',
            'autocomplete_light/select2.js',
            # Provide an additional i18 js.
            'autocomplete_light/vendor/select2/dist/js/i18n/ru.js',  
        )
        def build_attrs(self, *args, **kwargs):
            attrs = super(Select2BootstrapWidgetMixin, self).build_attrs(*args, **kwargs)
            attrs.setdefault('data-locale', 'ru')
            return attrs

class Select2Ru(Select2RuWidgetMixin, Select):
    pass


class Select2RuMultiple(Select2RuWidgetMixin, SelectMultiple):
    pass


class ModelSelect2Ru(QuerySetSelectMixin,
                     Select2RuWidgetMixin,
                     forms.Select):
    pass

class ModelSelect2RuMultiple(QuerySetSelectMixin,
                             Select2RuWidgetMixin,
                             forms.SelectMultiple):
    pass
```

Making it fully i18zable is a bit more complicated. You should include all i18n JS files to `Media.js`. Then in `build_attrs` you can get current locale from Django and set the `data-locale` to it.

sandino:
@gagarski Thanks for your help!

But I wonder then why would DAL have that i18n files in its code if they are not supposed to be used out of the box without going dirty ways.

gagarski:
> But I wonder then why would DAL have that i18n files in its code if they are not supposed to be used out of the box without going dirty ways.

I18n come from Select2 repo which is used in DAL as a submodule.

[...]

PR #746, opened 2016-09-21
Select2WidgetMixin Media as a dynamic property

luzfcb:
partially fixes the  issue #731.

corrects partly because, django `get_current_language` templatetag return a low case string with language code, and this value is commonly used on `html` tag, like `<html lang="pt-br">`, [see](https://github.com/django/django/blob/master/django/contrib/admin/templates/admin/base.html#L2-L3).  if html have any html tag with `lang="boo-bar-lang"` exists, select2 use the value to try load the language files.

The problem occur when the language is `pt-br`, `sr-cyrl`, `zh-cn`, `zh-tw`, because to these languages, some [i18n files of select2 is not low case](https://github.com/select2/select2/tree/master/src/js/select2/i18n), then select2 can not find the files. (`pt-BR.js`, `sr-Cyrl.js`, `zh-CN.js`, `zh-TW.js`)

### DIFF
--- a/src/dal_select2/widgets.py
+++ b/src/dal_select2/widgets.py
@@ -9,28 +9,40 @@ from dal.widgets import (
 
 from django import forms
 from django.utils import six
+from django.utils import translation
+from django.conf import settings
 
 
 class Select2WidgetMixin(object):
     """Mixin for Select2 widgets."""
 
-    class Media:
+    def _media(self):
         """Automatically include static files for the admin."""
+        # get_language can be return None
+        lang_code = translation.get_language()
+        if lang_code:
+            lang_code = translation.to_locale(lang_code).replace('_', '-')
+        else:
+            lang_code = ''
 
-        css = {
-            'all': (
-                'autocomplete_light/vendor/select2/dist/css/select2.css',
-                'autocomplete_light/select2.css',
-            )
-        }
+        _min = '' if settings.DEBUG else 'min.'
+
+        css = (
+            'autocomplete_light/vendor/select2/dist/css/select2.{}css'.format(_min),
+            'autocomplete_light/select2.css',
+        )
         js = (
             'autocomplete_light/jquery.init.js',
             'autocomplete_light/autocomplete.init.js',
-            'autocomplete_light/vendor/select2/dist/js/select2.full.js',
+            'autocomplete_light/vendor/select2/dist/js/select2.full.{}js'.format(_min),
+            'autocomplete_light/vendor/select2/dist/js/i18n/{}.js'.format(lang_code),
             'autocomplete_light/forward.js',
             'autocomplete_light/select2.js',
         )
 
+        return forms.Media(css={'all': css}, js=js)
+
+    media = property(_media)
     autocomplete_function = 'select2'
 
 
--- a/src/dal_select2/widgets.py
+++ b/src/dal_select2/widgets.py
@@ -16,26 +16,41 @@ from django.conf import settings
 class Select2WidgetMixin(object):
     """Mixin for Select2 widgets."""
 
-    def _media(self):
-        """Automatically include static files for the admin."""
+    def build_attrs(self, *args, **kwargs):
+        attrs = super(Select2WidgetMixin, self).build_attrs(*args, **kwargs)
+        lang_code = self._get_language_code()
+        if lang_code:
+            attrs.setdefault('data-locale', lang_code)
+        return attrs
+
+    def _get_language_code(self):
         # get_language can be return None
         lang_code = translation.get_language()
         if lang_code:
             lang_code = translation.to_locale(lang_code).replace('_', '-')
-        else:
-            lang_code = ''
+        return lang_code
 
+    def _media(self):
+        """Automatically include static files for the admin."""
         _min = '' if settings.DEBUG else 'min.'
+        i18n_file = ()
+        lang_code = self._get_language_code()
 
+        if lang_code:
+            i18n_file = (
+                'autocomplete_light/vendor/select2/dist/js/i18n/{}.js'.format(
+                    lang_code),
+            )
         css = (
-            'autocomplete_light/vendor/select2/dist/css/select2.{}css'.format(_min),
+            'autocomplete_light/vendor/select2/dist/css/select2.{}css'.format(
+                _min),
             'autocomplete_light/select2.css',
         )
-        js = (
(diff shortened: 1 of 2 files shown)


luzfcb:
@jpic It would be nice to slightly increase the maximum characters on a line.
django use 119

https://github.com/django/django/blob/master/setup.cfg#L8

https://github.com/django/django/blob/master/.editorconfig#L15

gagarski:
Should not we set `data-locale` HTML-attribute to enable selected locale?
```python
    def build_attrs(self, *args, **kwargs):
        attrs = super(Select2RuWidgetMixin, self).build_attrs(*args, **kwargs)
        attrs.setdefault('data-locale', 'ru')
        return attrs
```

gagarski:
> The problem occur when the language is pt-br, sr-cyrl, zh-cn, zh-tw, because to these languages, some i18n files of select2 is not low case, then select2 can not find the files. (pt-BR.js, sr-Cyrl.js, zh-CN.js, zh-TW.js)

The only universal way to handle it is to store a mapping from Django locale to Select2 locale and fall back to no locale if there is no matching Select2 locale

src/dal_select2/widgets.py
```diff
@@ -9,28 +9,40 @@
...
+        css = (
+            'autocomplete_light/vendor/select2/dist/css/select2.{}css'.format(_min),
+            'autocomplete_light/select2.css',
+        )
         js = (
             'autocomplete_light/jquery.init.js',
             'autocomplete_light/autocomplete.init.js',
-            'autocomplete_light/vendor/select2/dist/js/select2.full.js',
+            'autocomplete_light/vendor/select2/dist/js/select2.full.{}js'.format(_min),
+            'autocomplete_light/vendor/select2/dist/js/i18n/{}.js'.format(lang_code),
```
gagarski: What happens if `lang_code == ''` here?

luzfcb:
@gagarski done.

jpic:
> The problem occur when the language is pt-br, sr-cyrl, zh-cn, zh-tw, because to these languages, some i18n files of select2 is not low case, then select2 can not find the files. (pt-BR.js, sr-Cyrl.js, zh-CN.js, zh-TW.js)

Could we perhaps iterate over the translation files in select2's locale directory and map them to django locales ?

luzfcb:
@jpic hard disk read operation seems to me can be slow. since it will run every request. must leave. I am going into class

jpic:
Yes we shouldn't check the filesystem at every request, but we could have a memoizing class function, ie.:

    locale_map = None
    @classmethod
    def get_select2_locale(cls, django_locale):
        if locale_map is None:
            # walk the filesystem, set cls.locale_map
        return cls.locale_map[django_locale]

Would this prevent un-needed FS reads ?

gagarski:
@jpic Is there a Django-way to iterate over some static directory files? I do not know any.

gagarski:
@luzfcb As far as I can see from django docs, this WILL work for `pt-br`.
`translation.to_locale("pt-br")` will return `"pt_BR"` which you convert to `"pt-BR"`.

jpic:
> @jpic Is there a Django-way to iterate over some static directory? I do not know any.

Didn't find any in the source code.

Another way is to just add the 4 exceptions in a list, it'll need maintenance but not so much after all.

gagarski:
@jpic Seems like there are much more than 4 exceptions.
Here is a Django locales list: https://github.com/django/django/tree/master/django/conf/locale
And here is one from Select2: https://github.com/select2/select2/tree/master/dist/js/i18n

So here are some examples (Django -> Select2):
```
es_AR -> es
es_CO -> es
es_MX -> es
es_NI -> es
es_PR -> es
en_AU -> en
en_GB -> en

# More fun!
sr -> sr-Cyrl
sr_Latn -> sr

zh_Hans -> ???
zh_Hant -> ???

??? -> zh-CN
??? -> zh-TW
```

Also, some Django locales are missing from Select2. Of course we can add static JS file anyway but this will cause 404 errors for unknown locale static files.
Looks like the whole map of locales should be added and maintained :(

luzfcb:
https://groups.google.com/forum/#!searchin/select2/lang|sort:relevance/select2/jKBjwqtBbf0/sndL1YQ3CAAJ

luzfcb:
i rename "data-locale" to "data-autocomplete-light-language", to maintain the consistency and add `language` parameter to select2 start function

jpic:
Yes, yes, yesssss

Keep up the great work 

:guitar:

jpic:
So, should we merge it as is or should we finish it up first ? I suppose you want pt_BR ?

marcossantiago:
Hey guys, 

This commit would solve my problem. Any chance of merging it to master anytime soon ?

Thanks for you work.

jpic:
Well it's pretty good as-is but given that @luzfcb is located in Brazil I was wondering if he wanted to add pt_BR support first.

src/dal_select2/widgets.py
```diff
@@ -9,28 +9,55 @@
...
 
-        css = {
-            'all': (
-                'autocomplete_light/vendor/select2/dist/css/select2.css',
-                'autocomplete_light/select2.css',
+    def _media(self):
+        """Automatically include static files for the admin."""
+        _min = '' if settings.DEBUG else 'min.'
+        i18n_file = ()
+        lang_code = self._get_language_code()
```
Pavel-Y: For Russian language, `lang_code` here has value `'ru-RU'` and i18n file name is ru.js. So I had to replace this line with `lang_code = self._get_language_code().split('-',1)[0]` 
and then it worked for me. Otherwise, the file could not be loaded (error 404).

jpic: Yep, we should add a locale directory for that or find a generic replacement that uses a string translated in django itself (best).

jpic:
Anybody wants to pick this up ? CI will be green tomorrow it'll be the right time to try to merge features for a new release.

gagarski:
Big part of this problem is solved by #872. Now we have to map Django locales to Select2 ones (I don't see non-hacky way to do it without explicitly storing it in a dictionary).

jpic:
Then hacky it'll be until we figure better with @kevinbrown. Given that
django admin wants to integrate select2 I assume they will make something
up.

Le 24 juil. 2017 10:00 PM, "Kirill Gagarski" <notifications@github.com> a
écrit :

jpic:
Integrated in #954, will be in next release

PR #753, opened 2016-09-25
Docs dal_queryset_sequence proposal.

chubz:
I've put these simple notes here, maybe we should expand those further. I hope you give further input.
Maybe in install there should be a list of DAL apps that can be installed. Then in each part of DOCS if there is a reference to that app just put a note saying it need to be added to installed apps.


jpic:
Thanks !

chubz:
No problem, glad I could help!

PR #770, opened 2016-11-18
Select2 updated to 4.0.3

gagarski:


ibnbay99:
Js Select2 4.0.3 is release from May 2016, why just recently this updated on django-autocomplete-light? (sorry not mean to be harsh) :)

Is there another option to update or get own file reference like define on django settings the path for latest select2 plugin to override internal static select2 on django-autocomplete-light?

gagarski:
You can create your own widget classes derived from DAL ones and override `Media` nested class. Another option is to put Select2 js to project or your own app `static` directory having the same path (`/path/to/your/project/yourapp/static/autocomplete_light/vendor/...`) and play around with app order and static files loaders.

jpic:
Nice one @gagarski, I just tested with tox and BROWSER=chrome on Arch Linux with this patch:

```
iff --git a/tox.ini b/tox.ini
index 8b078fa..224aa5d 100644
--- a/tox.ini
+++ b/tox.ini
@@ -26,7 +26,7 @@ deps =
     mysql: mysql-python
     -rtest_project/requirements.txt

-passenv = DISPLAY XAUTHORITY XDG_* PIP_*
+passenv = DISPLAY XAUTHORITY XDG_* PIP_* BROWSER
```

Merge when you're ready to follow-up ;)

:guitar:

PR #773, opened 2016-11-19
Fix jQuery not sending X-CSRFToken header

### LINKED ISSUE #772: CSRF verification failed when trying to create new options from Select2
I have a Select2 field on a ModelForm. This field is supposed to allow users to add new options but I always get `CSRF verification failed. Request aborted.` from the view that was supposed to help create the new object, and I get `Forbidden (CSRF token missing or incorrect.): /movement-autocomplete/` in the console logs.

Inspecting the HTTP POST request, I can see that its "x-csrftoken" header is set to "null", which I suppose explains the error.

This only happens when DEBUG is False, otherwise it works perfectly.

Is this a known issue that turns out to be something subtle that people miss? Any ideas on what could be wrong?

I'm setting up the field like this:

```Python
widgets = {
    'movement': autocomplete.ModelSelect2(
        url='movement-autocomplete',
        attrs={'data-placeholder': 'Movement'}
    ),
}
```

The view that feeds the field is the following:

```Python
@method_decorator(login_required, name='dispatch')
class MovementAutocomplete(autocomplete.Select2QuerySetView):
    create_field = 'name'

    def has_add_permission(self, request):
        """Allow any signed in user to add a new movement."""
        if not request.user.is_authenticated():
            return False

        return True

    def get_queryset(self):
        qs = Movement.objects.all()

        if self.q:
            qs = qs.filter(name__istartswith=self.q)

        return qs
```

borfast:
Apparently jQuery doesn't add the X-CSRFToken header when xhr.setRequestHeader() is used. $.ajax has a `headers` option which works. Fixes #772.

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -73,8 +73,8 @@
                     text: data.id,
                     forward: get_forwards($(this))
                 },
-                beforeSend: function(xhr, settings) {
-                    xhr.setRequestHeader("X-CSRFToken", document.csrftoken);
+                headers: {
+                    'X-CSRFToken': document.csrftoken
                 },
                 success: function(data, textStatus, jqXHR ) {
                     select.append(


borfast:
I assume the package maintainers know this but the Travis errors are due to a few test-specific things, not anything related to this PR.

borfast:
I don't drink but if I did, I'd say I must have been drunk when I submitted this PR...

Please ignore it. The issue still stands. I'll investigate further and try to figure out what's going on.

PR #795, opened 2017-01-25
Fix RemovedInDjango20Warning for 'reverse' imports

luksen:
Hi,

I only did a search and replace for this but I don't see what could go wrong^(tm).

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -85,7 +85,7 @@ Create a :django:label:`named url<naming-url-patterns>` for the view, ie:
 Ensure that the url can be reversed, ie::
 
     ./manage.py shell
-    In [1]: from django.core.urlresolvers import reverse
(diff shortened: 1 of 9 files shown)


jpic:
Seems like it breaks Django 1.8 support:
https://travis-ci.org/yourlabs/django-autocomplete-light/jobs/195213265​

Note sure up to when we need to support this version, but that migth be
until the end of this year.

Meanwhile, we could have try/except ImportError :)

luksen:
That's how other projects seem to handle it as well. I reverted the tutorial back to the old import because I think the whole try-except would be ugly there.

jpic:
Nice ! I think the best in the docs is to have the new import though,
perhaps the old one could be left in a comment.

luksen:
Good idea! Something like this?

PR #798, opened 2017-02-03
Exclude src/dal/test instead of src/dal/tests to fix flake8 failure

thauk-copperleaf:


thauk-copperleaf:
@jpic This should fix the checkqa build (flake8), at least

jpic:
I don't understand, what does this fix exactly ?

thauk-copperleaf:
https://travis-ci.org/yourlabs/django-autocomplete-light/jobs/196763240

```
...
checkqa runtests: commands[0] | flake8 --show-source --max-complexity=7 --ignore=D203 --exclude src/dal/autocomplete.py,tests src
src/dal/test/stories.py:48:1: C901 'BaseStory.find_option' is too complex (8)
    def find_option(self, text, click=False):
^
ERROR: InvocationError: '/home/travis/build/yourlabs/django-autocomplete-light/.tox/checkqa/bin/flake8 --show-source --max-complexity=7 --ignore=D203 --exclude src/dal/autocomplete.py,tests src'
...
```

It seemed to me that you were attempting to exclude tests from flake8. Am I incorrect?

jpic:
Right, it would be better to decomplexify that function but i'm not sure if it's really possible.

jpic:
> It seemed to me that you were attempting to exclude tests from flake8. Am I incorrect?

Nope, all code should be covered by flake8, just with different set of options.

PR #799, opened 2017-02-06
Support serializing UUIDs and add tests for models with UUIDs as PKs

### LINKED ISSUE #781: JSON not serializable when using UUIDField
When using an UUIDField for the ID (PK) field, the ID is not directly serializable, since it is of type UUID. The function `get_result_value` in the class `BaseQuerySetView` in `views.py` should take this into account.
I suggest changing this:
```
def get_result_value(self, result):
    """Return the value of a result."""
    return result.pk
```
to `return six.text_type(result.pk)` as you do it for the label.

blag:
~~I just copied all of the tests from `select2_foreign_key` into `select2_foreign_key_uuid`~~

Closes #781.

### DIFF
--- a/src/dal/views.py
+++ b/src/dal/views.py
@@ -55,7 +55,7 @@ class BaseQuerySetView(ViewMixin, BaseListView):
 
     def get_result_value(self, result):
         """Return the value of a result."""
-        return result.pk
+        return str(result.pk)
 
     def get_result_label(self, result):
         """Return the label of a result."""


jpic:
Are you sure we need functional tests for this ? Or could we get away with only unit tests ?

:guitar:

blag:
We probably just need unit tests.

However, I am pressed for time and I need this to work with UUIDs, so I just copied and pasted `select2_foreign_key` and appended `_uuid` where appropriate.

Suggestions on what/where to remove are very welcome! :smile:

jpic:
Can remove the functional testing code you have copied ? Just leave the single line you want to change for now.

I'm sure someone will be happy to find an easy contribution such as adding a test for this change:

```
class BaseQuerySetView(ViewMixin, BaseListView):
￼      def get_result_value(self, result):
￼          """Return the value of a result."""
￼ -        return result.pk
￼ +        return six.text_type(result.pk)
￼ ```

Ie. in src/dal/tests/test_views.py

File: src/dal/views.py [hunk 1]
```diff
@@ -55,6 +55,7 @@ def has_more(self, context):
 
     def get_result_value(self, result):
         """Return the value of a result."""
+        # return six.text_type(result.pk)
```
jpic: What happened ?

blag: [Test 199524496](https://travis-ci.org/yourlabs/django-autocomplete-light/builds/199524496) is failing on Django 1.10 *without* my change.

For future reference, the diff for this test run is this:

```diff
commit 1dd93245bf8e44fdd9a3a05ca7a30082f0e2794c
Author: blag
Date:   Mon Feb 6 12:23:04 2017 -0600

    Support serializing UUIDs - closes #781

diff --git a/src/dal/views.py b/src/dal/views.py
index b23aa5c..57824dd 100644
--- a/src/dal/views.py
+++ b/src/dal/views.py
@@ -55,6 +55,7 @@ class BaseQuerySetView(ViewMixin, BaseListView):

     def get_result_value(self, result):
         """Return the value of a result."""
+        # return six.text_type(result.pk)
         return result.pk

     def get_result_label(self, result):
```

blag:
The Travis build seem to be failing intermittently. I don't think the current failures are due to this change.

jpic:
It seems consistently failing on dj1.10 but not because of this change. However, can we have a new test against regression ?

src/dal/views.py
jpic: Right, don't worry about it, try to add your own regression test for this use case ;)

blag:
@jpic I'm happy to add a regression test for this, but you'll have to guide me a bit. The `test_project` directory is a little intimidating. Where should I add the regression test? What should the test look like/do? I had previously copy/pasted the `select2_foreign_key` tests and that was overkill.

jpic:
I think just a new test in src/dal/test_view.py would do the trick, just to check that this method returns the right thing under both circumstances.

blag:
@jpic I don't see that file anywhere in the `master` branch. Is there another branch I should use?

jpic:
Go ahead and create it !

jpic:
Could you rebase on master please ? It's currently passing tests, i'd like to merge this in the next release.

blag:
@jpic Rebased. 😄

blag:
@jpic Can you give me a rough ETA for the next release that will include this? I'm trying to figure out if I will need to deploy with my customized version or if I will be able to pip install yourlabs' version. Thanks!

jpic:
Releasing ASAP is fine, but I'm more worried about your release process. Are you planning into re-installing the same package and dependencies over and over again every time you deploy your project ? Like, in virtualenvs in production ?

blag:
Inside a docker container, amongst other things, I use pip to install everything in a `requirements.txt` file. So I can either specify it in that or I can keep my customized version in my git repo and use that instead.

Why? Is there a better way than that?

jpic:
Nope, that's the best afaik :guitar: 

I was just wondering why you needed a release ETA

PR #807, opened 2017-03-09
Fix documentation error where ListSelect2 widget is named incorrectly

FuzzAU:
I was trying to follow the tutorial to create an autocomplete with a static list rather than a QuerySet and was getting errors referring to Select2List. It turns out that isn't what it's called.

This commit just updates the documentation


jpic:
Nice, thanks !

PR #813, opened 2017-03-17
Return 400 on invalid input

EvaSDK:
A simple patch to return proper information to the client side that its input data is invalid.
This is especially useful to silence errors that are not server-side when coupled with reporting tools such as [sentry](https://sentry.io/welcome/), [Airbrake](https://airbrake.io/), etc.

I did not add unittests as I got lost in the structure for functional tests as library specific tests but if you tell me where to look, I'll add them.

### DIFF
--- a/src/dal/views.py
+++ b/src/dal/views.py
@@ -5,12 +5,16 @@ import json
 from django import http
 from django.contrib.auth import get_permission_codename
 from django.core.exceptions import ImproperlyConfigured
+from django.http import HttpResponseBadRequest, HttpResponseNotAllowed
 from django.utils import six
 from django.views.generic.list import BaseListView
 
 
 class ViewMixin(object):
     """Common methods for autocomplete views.
+
+    It is assumed this view will be used in conjunction with a Django
+    :py:class:`View` based class that will that will implement OPTIONS.
 
     .. py:attribute:: forwarded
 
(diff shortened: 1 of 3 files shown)


src/dal/views.py
```diff
@@ -25,9 +26,16 @@ class ViewMixin(object):
...
-            self.forwarded = json.loads(request.POST.get('forward', '{}'))
+        try:
+            self.forwarded = json.loads(request.GET.get('forward', '{}'))
+            if request.method == 'POST':
+                self.forwarded = json.loads(request.POST.get('forward', '{}'))
+
+            if not isinstance(self.forwarded, dict):
+                raise ValueError
+        except ValueError:
+            return HttpResponseBadRequest('Invalid JSON data')
```
jpic: While you're at it, would you like to remove some SLOCS with:

    self.forwarded = getattr(request, 'POST' if request.method == 'POST' else 'GET').get('forward', '{}')

Also, it think the code would be a better read if the try block only contains the loads() call, and then check if it's a dict and return 400 otherwise: it seems like it would be easier to read and understand with two distinct return HttpResponseBadRequest() statements rather that a raise ValueError/except ValueError.

EvaSDK: I was playing with a fix according to your comment but then I figured that only GET and POST are supported here so maybe we want to return a `405 Method Not Allowed` if the method is not supported and that would mean implementing `OPTION` as well and then this pretty much resemble the Django default `View`. Should I modify this view to subclass it directly ?

jpic: 405 sounds useful, OPTIONS not so much, but if you're up for it go ahead, don't forget we'll need unit tests so you might not want to go through implementing OPTIONS for this reason.

EvaSDK:
Should be all good now. I left OPTIONS out as all cases I looked at ended up in View being in the inheritance tree so it will be handled automatically in most cases. The view now duplicates the View class check for 405 but I think there is no way around it.

jpic:
Released in 3.2.4, grats !

Issue #783, opened 2016-12-24
search_fields attribute / property


toudi:
Hello. I was looking trough the source code and the tutorial / documentation.

Here's the code from documentation, when it comes to registering an autocomplete view:

```
class CountryAutocomplete(autocomplete.Select2QuerySetView):
    def get_queryset(self):
        # Don't forget to filter out results depending on the visitor !
        if not self.request.user.is_authenticated():
            return Country.objects.none()

        qs = Country.objects.all()

        if self.q:
            qs = qs.filter(name__istartswith=self.q)

        return qs
```

However, I was wondering whether we could keep simple things simple. Namely, when we register the autocomplete view, like so:

```
urlpatterns = [
    url(
        r'^country-autocomplete/$',
        CountryAutocomplete.as_view(),
        name='country-autocomplete',
    ),
]
```

could we have a simpler syntax, something like..

```
urlpatterns = [
    url(
        r'^country-autocomplete/$',
        name='country-autocomplete',
        model=Country,
        search_fields=('foo', 'bar__baz'),
    ),
]
```

then of course the underlying logic would construct a Q() object:

```
qs = super( something, self).get_queryset()
if self.q:
    filters = Q()
    for field in self.search_fields:
        filters |= Q(**{'%s__icontains' % field: self.q})
    qs = qs.filter(filters)
```

What I mean is that I totally get the point of doing a generic view when there is complex logic involved (like creating objects, complex / custom filtering, etc, etc, etc). However dal can also be used to very simple tasks and this whole block of code (extending a class) contains a single method to handle simple filtering - seems to be an overkill.

If you would agree to such changeset, I would be more than happy to provide a PR

my kindest regards,

jpic:
Hi Mateusz !

Did you mean to show an example like this ?

```
urlpatterns = [
    url(
        r'^country-autocomplete/$',
        name='country-autocomplete',
        dal.ModelSelect2Autocomplete(
            model=Country,
            search_fields=('foo', 'bar__baz'),
        )
    ),
]
```

I really like this, but I think any kind of search_fields logic should be maintained in a search app rather than in django-autocomplete-light, is there any way we could reuse [django-search-views](https://github.com/inmagik/django-search-views) ? Here's an example in their README which shows they already have some kind of query building logic: 

```
from search_views.search import SearchListView
from search_views.filters import BaseFilter

class ActorsFilter(BaseFilter):
    search_fields = {
        'search_text' : ['name', 'surname'],
[...]

PR #815, opened 2017-03-17
Simplify customization of autocomplete views

EvaSDK:
Make it easier to select the adequate field to filter against so subclasses can properly use super() and stop having to redefine whole queryset because of this.

Hardly a big change but this annoyed me pretty much when I realized I had to write more lines than needed for a simple attribute name change.

### DIFF
--- a/src/dal/views.py
+++ b/src/dal/views.py
@@ -43,10 +43,15 @@ class BaseQuerySetView(ViewMixin, BaseListView):
         find any value matching "foo". When the user does click 'Create "foo"',
         the autocomplete script should POST to this view to create the object
         and get back the newly created object id.
+
+    .. py:attribute:: model_field_name
+
+        Name of the Model field to run filter against.
     """
 
     paginate_by = 10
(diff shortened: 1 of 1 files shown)


jpic:
We'll need tests and documentation, here's some background history that you might be able to use: https://github.com/yourlabs/django-autocomplete-light/issues/783

jpic:
Hello ! Would it be possible to add just a small unit test before we release this ?

EvaSDK:
I'd rather delay the inclusion of this. I have had a look at the history as suggested and I started work on it. Some work priorities got in the way though hence the delay.

toudi:
damn, I am so utterly sorry that I failed to deliver this feature in half a year.. :( I just managed to find some free time today and I'll still see what can I come up with

jpic:
@toudi don't worry about it, time is not the same, for humans and for software, time doesn't matter only <3 matters :dancer:

jpic:
Added in #954

PR #816, opened 2017-03-21
App added to testproject to render two autocomplete forms 

shubhamdipt:
Migration files added
Settings changed to run it at one go
Added an app to testproject (select2_outside_admin_multiple)
Edited base.html to include link to the new app.


jpic:
Nice ! Thanks for your contribution B)

Why are there so many new models ? Can't we just add one new app and use relations to 'self' to test autocompletes ?

shubhamdipt:
Yup I agree ... it was just a quick copy paste stuff  :P to replicate the issue. By the way it makes it a more realistic situation though :)

jpic:
I think it will be easier if we start over, there's way to much code. Would it be possible instead to just make a new view in the new app and reuse everything from select2_outside_admin in there ? ie. render the form from the other app twice, instead of defining a new one. This will also help documentation !

jpic:
You're a great programer but I don't see the point for this and we already got many apps with tests to maintain

Issue #819, opened 2017-03-22
autocomplete fields not initialized when DOM is ready before select2.js is loaded


apinsard:
When the DOM is ready before `select2.js` script is loaded, the `autocompleteLightInitialize` event is sent while the listener is not yet set.

This results in autocomplete fields not initialized.

I worked this around by calling `$('[data-autocomplete-light-function]:not(.select2-hidden-accessible)').trigger('autocompleteLightInitialize')` in a script called after `select2.js`.

Is there a more proper way to deal with this ?

jpic:
Interresting, have you tried moving select2.full.js at the top of the widget media list ? https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal_select2/widgets.py#L26

apinsard:
Actually, we use RequireJS to ensure scripts dependencies. So I set the dependencies as the following:

- jquery.init.js depends on jquery
- autocomplete.init.js depends on jquery.init.js
- select2.full.js depends on autocomplete.init.js
- forward.js depends on select2.full.js
- select2.js depends on forward.js.

I tried the following, according to your comment:

- jquery.init.js depends on jquery
- select2.full.js depends on jquery.init.js
- autocomplete.init.js depends on select2.full.js
- forward.js depends on autocomplete.init.js
- select2.js depends on forward.js.

But this didn't change anything.

To me, the issue lies in the fact that [`autocompleteLightInitialize` event is triggered](https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal/static/autocomplete_light/autocomplete.init.js#L45) when the [DOM is ready](https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal/static/autocomplete_light/autocomplete.init.js#L49-L50). Which may be before [the listener is set up](https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal_select2/static/autocomplete_light/select2.js#L3) unless you call your scripts synchronously at the end of the `<body>` which I don't want to do.

Forcing `select2.js` to be executed before `autocomplete.init.js` with the following precedence:

- jquery.init.js depends on jquery
- select2.full.js depends on jquery.init.js
- select2.js depends on select2.full.js.
- autocomplete.init.js depends on select2.js
- forward.js depends on autocomplete.init.js

[...]

PR #820, opened 2017-03-22
Fixed JS autocomplete initializer (#819)

apinsard:
Call initialize() manually once 'autocompleteLightInitialize' event
listener is set on select2.js.

This allows the initializer to be called even when the DOM is ready
before select2.js was loaded.

Also replaced 'element = $(this).get(0)' by 'element = this' which
should be equivalent in this scope.


jpic:
Thanks !

Could you rebase this on master so I can check if this broke any tests (which are now passing).

In the best case, i'll be able to merge and release this this week end.

jpic:
Hello @apinsard !

Tests are passing on master, could you rebase please ?

When this is rebased and doesn't break tests I'd like to release this !

apinsard:
@jpic Done

jpic:
It didn't pass any integration test on Travis-ci, does it pass for you ?

Merci Antoine <3

apinsard:
Indeed... Yet I don't get what could make tests fail as I haven't changed any python file

jpic:
Apparently no selenium test is passing anymore.

jpic:
I still have hope in this !

Salutations from POITOU CHARENTE <3

PR #821, opened 2017-03-24
Made Select2ListView return case insensitive results by default

DSynergy:
-Made select2ListView return a case insensitive match by default
to match the select2QueryView
-Added a kwarg to toggle case sensitivity in select2ListView get()

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -556,7 +556,8 @@ create a ListSelect2 widget with:
     widget = autocomplete.ListSelect2(url='country-list-autocomplete')
 
 With this in place, if a user types the letter ``f``' in the widget, choices
-'France', 'Fiji', and 'Finland' would be offered.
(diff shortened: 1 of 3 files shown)


jpic:
I think we should better drop case_insensitive and just leave the insensitive search.

However, we need to add a regression test.

DSynergy:
Sure I can drop the kwarg, that is no problem. I actually work at eGauge systems with David and Joshua, the original contributors for the list select, so just patching their work.

What sort of regression test would you want?

Should the Travis-cl be passing currently? Looks like it is missing a package 'white-noise'?

jpic:
Just a test for your patch, feel free to add more if you're into this kind of stuff :dancer:

DSynergy:
Okay, I have updated the commit and added a thorough test for both my change and the Select2ListView in general

jpic:
Really awesome, thanks !!!

DSynergy:
Sounds good. Glad I could help

PR #822, opened 2017-03-25
case insensitive Select2ListView

jsurloppe:
Hi,

This PR allow to do case insensitive research in a Select2ListView.

### DIFF
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -62,17 +62,22 @@ class Select2QuerySetView(Select2ViewMixin, BaseQuerySetView):
 
 class Select2ListView(ViewMixin, View):
     """Autocomplete from a list of items rather than a QuerySet."""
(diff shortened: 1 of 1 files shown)


jpic:
Does case sensitive search really make sufficient sense to keep it at all ?

jsurloppe:
I don't need it myself, i  have  done it like this for ensuring backward compatibility.

jpic:
Ok let's just get rid of case sensitive search and only do case insensitive instead would that work for you then ?

DSynergy:
Funny...I just did this in #821

jsurloppe:
LGTM, i suggest merging #821, closing #822 and i'll modify #823 according.

jpic:
OMG I'm so tired, I thought it was the same PR ... We need to recruit core devs, I can barely follow up.

PR #823, opened 2017-03-26
Optgroup list

jsurloppe:
This PR needs #822 and introduces a select2 style optgroup list view.

It didn't handle creation  as AFAIK there is no standard way to indicate that item should be created under which group.

There's still a problem with the previous selected choice that didn't appear selected in the admin, but that was already a problem with the original Select2ListView. I will try to fix that in another PR if a find the time.


src/dal_select2/views.py
```diff
@@ -106,3 +113,46 @@ def post(self, request):
...
+            results_dict.setdefault(group, [])
+            results_dict[group].append(value)
+
+        return http.HttpResponse(json.dumps({
+            'results':
+                [{"id": x, "text": x} for x in results_dict.pop(None, [])] +
+                [{"id": g, "text": g, "children": [{"id": x, "text": x}
+                                                   for x in l]}
+                 for g, l in six.iteritems(results_dict)]
+        }))
```
jpic: Hehe that's pretty cool mate, however, we need a test and documentation, otherwise put it in a separate file such as hacks.py so at least we have it for the record until somebody wants to rewrite it in tdd :guitar:

jpic:
Hi ! Tests are passing on travis now, could you rebase this on the current master please ?

Do you think you add a minimal example in test_project ?

Thanks !

jsurloppe:
Rebased, tested and minimally documented. :)
I've squashed the previous branch for removing the other PR dep.

jpic:
Absolutely delicious patch <3

PR #827, opened 2017-04-03
Fix tests

jpic:
I can't believe it's about to work...


jpic:
So it seems like we're entering a dark period in terms of code quality. Supporting both django <1.10 and django >1.10 requires logic duplication ...

This is going to be the last release with all features in the core repo. I want to leave things that depend **only** on django in this repo, and move the rest which depends on third party apps such as tagging, taggit, gm2m, etc ... into a dal-extras repo so that we don't depend on their release cycle to follow up with django development, but also to make the core repo easier to maintain and contributions more fluid because we'll also be dispatching CI.

jpic:
Now to the point where everything passes in local (haven't tested with pypy though) but fail on travis with dj >= 1.10 .... @blueyed what do you think ?

jpic:
:tada:

jpic:
A review would be nice to have of course friends ! :birthday:

luzfcb:
@jpic LGTM.

src/dal/test/stories.py
```diff
@@ -331,7 +344,10 @@ def get_labels(self):
             self.get_field_labels_selector()
         )
 
-        return [six.text_type(label.text) for label in labels]
+        return [
+            six.text_type(label.text).replace('\xd7', '')
```
blueyed: What is `\xd7`?  Could need a comment.

test_project/settings/base.py
```diff
@@ -1,7 +1,11 @@
 import os
+import sys
 import django
 
-DEBUG = os.environ.get('DEBUG', False)
+DEBUG = os.environ.get(
+    'DEBUG',
+    'py.test' in sys.argv[0] or 'runserver' in sys.argv[0]
```
blueyed: Should also catch/prefer `pytest` (without the dot).

jpic:
test_filter_options seems to still block from time to time, i suspect it's the refresh_autocomplete method that needs some checks and waits to be added, i'm driving during the next 8-10 hours so if anybody has the chance to take a look it would be neat :P meanwhile let's merge this

PR #833, opened 2017-04-15
rename_forward tests

gagarski:
I have added `test_functional` suite that reuses scenarios from `linked_data` app.

`test_forms` suite from `linked_data` seems useless for `rename_forward` app since it won't test any other functionality.

I am thinking of writing tests that checks if Select2 widget is rendered correctly, especially the forward configuration part (perhaps we can add something like that to `linked_data` app). The other good thing to test can be `LinkedDataView` to check if we're running functional tests with properly working view (the forward feature itself does not add any features to DAL views).

### DIFF
--- a/test_project/rename_forward/test_functional.py
+++ b/test_project/rename_forward/test_functional.py
@@ -1,9 +1,10 @@
-from .models import TModel
 import linked_data.test_functional
+
+from .models import TModel
 
 
 class RenameForwardAdminLinkedDataTestTest(
-    linked_data.test_functional.AdminLinkedDataTest):
+        linked_data.test_functional.AdminLinkedDataTest):
     """Reusing functional test scenarios from linked_data app."""
 
     model = TModel
--- a/test_project/rename_forward/test_functional.py
+++ b/test_project/rename_forward/test_functional.py
@@ -1,0 +1,10 @@
+from .models import TModel
+import linked_data.test_functional
+
+
+class RenameForwardAdminLinkedDataTestTest(
+    linked_data.test_functional.AdminLinkedDataTest):
+    """Reusing functional test scenarios from linked_data app."""
+
+    model = TModel
+    inline_related_name = 'inline_test_models_rf'


jpic:
I can't find a way to assert that a particular url was used with an AJAX request with the python/selenium we have. Perhaps the only way to test it is to make the view only returns choices if a particular field was renamed and forwarded completely.

You don't need the other tests from linked data I think. It looks like just a scenario like:

- open rename_forward admin form,
- fill in some field that will be forwarded+renamed,
- type something in the autocomplete,
- have the view return something expected if the forward+rename worked,
- assert that the select2 widget has the right value.

What do you think ? That's just one functional test but I think it would cover the rename feature for the next year(s).

gagarski:
> You don't need the other tests from linked data I think. It looks like just a scenario like:
> 
> - open rename_forward admin form,
> - fill in some field that will be forwarded+renamed,
> - type something in the autocomplete,
> - have the view return something expected if the forward+rename worked,
> - assert that the select2 widget has the right value.

This is exactly what `linked_data.test_functional.AdminLinkedDataTest.test_filter_options` do. And this scenario works for `rename_forward`. Other scenarios test the same functionality in inline forms and I think that these scenarios can be actual in `rename_forward` (something can go wrong with interferenction of `get_forwards()` and inline forms JS).

jpic:
Do you mean that this test should fail if we remove the JS bit that does
the rename ?

Le 25 avr. 2017 8:38 PM, "Kirill Gagarski" <notifications@github.com> a
écrit :

gagarski:
If `owner` is not renamed to `possessor` by JS and 42 is not forwarded as
`secret`, then these tests will fail.

gagarski:
> if we remove the JS bit that does
the rename ?

BTW, currently it's hard to imagine because non-renaming forward now is a particular case of renaming forward. Renaming forward is "Take value from `owner` field and pass it to DAL view with key `possessor`". Non-renaming is "Take value from `owner` field and pass it to DAL view with key `owner`"

jpic:
Yesss ! In this case of course please by all means merge this B)

I'm trying to test it as i'm speaking but I got some unrelated errors.

gagarski:
@jpic I am still thinking of adding python tests which ensure that widget is rendered correctly, especially forward configuration (and maybe one like this in `linked_data`).

jpic:
That would be surely good value for DAL, but this PR already covers your changes so perhaps you want it merged rather sooner than later ?

PR #836, opened 2017-04-20
Queryset sequence view to display actual model name

jsoa:
The change set here uses the actual model, instead of the QuerySetSequence proxy model.

Before
qss-before

After
qss-after

### DIFF
--- a/src/dal_queryset_sequence/views.py
+++ b/src/dal_queryset_sequence/views.py
@@ -52,7 +52,7 @@ class BaseQuerySetSequenceView(BaseQuerySetView):
                           result.pk)
 
     def get_model_name(self, model):
-        """Return the name of the model, fetch the parent if model is a proxy"""
+        """Return the name of the model, fetch parent if model is a proxy"""
         if model._meta.proxy:
             try:
                 model = list(model._meta.parents.keys())[0]
--- a/src/dal_queryset_sequence/views.py
+++ b/src/dal_queryset_sequence/views.py
@@ -50,3 +50,12 @@ class BaseQuerySetSequenceView(BaseQuerySetView):
         """Return ctypeid-objectid for result."""
         return '%s-%s' % (ContentType.objects.get_for_model(result).pk,
                           result.pk)
+
+    def get_model_name(self, model):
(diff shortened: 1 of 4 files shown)


src/dal_select2_queryset_sequence/views.py
```diff
@@ -43,9 +43,19 @@ def get_results(self, context):
             groups.setdefault(type(result), [])
             groups[type(result)].append(result)
 
+        def get_model_name(model):
+            # Fetch the parent model if this is a proxy
+            if model._meta.proxy:
+                try:
+                    # Fetch the proxied model instead
+                    model = model._meta.parents.keys()[0]
+                except (IndexError, ):
```
jpic: No need for parenthesis here

jpic:
Nice ! Shoul we have a unit test for get_model_name ?

I think it would help testing compat accross django versions.

jsoa:
Updated:

- Added test case
- Moved `get_model_name` to `BaseQuerySetSequenceView`
- Updated try (per comment)

src/dal_queryset_sequence/views.py
```diff
@@ -50,3 +50,12 @@ def get_result_value(self, result):
...
         return '%s-%s' % (ContentType.objects.get_for_model(result).pk,
                           result.pk)
+
+    def get_model_name(self, model):
+        """Return the name of the model, fetch the parent if model is a proxy"""
+        if model._meta.proxy:
+            try:
+                model = model._meta.parents.keys()[0]
+            except IndexError:
+                pass
```
jpic: Is it possible to test when there there is an IndexError ?

jpic:
Not sure why the checkqa test isn't passing anymore, I re-ran it on master here: https://travis-ci.org/yourlabs/django-autocomplete-light/jobs/222043382 if that one passes then checkqa should pass here too

Also, any idea why it doesn't seem to pass on python 3 ?

Keep up the gr8 work <3

jsoa:
Updated:

- Fixed the python 3 issue with `.keys()`
- Added index error test

jpic:
Gratz B)

Issue #806, opened 2017-03-07
Just when I use forward got the message "$(...).getFormPrefix is not a function"


Fabitosoft:
Hello,

I've been using this amazing library 3 month ago but today, when I tried to use forward in one o my fields there is a message that said "$(...).getFormPrefix is not a function". These are the two fields that I tried to use:
```

cliente_biable = forms.ModelChoiceField(
        queryset=Cliente.objects.all(),
        widget=autocomplete.ModelSelect2(url='biable:cliente-autocomplete'),
        required=False,
        label='Cliente CGuno'
    )

contacto = forms.ModelChoiceField(
        queryset=ContactoEmpresa.objects.all(),
        widget=autocomplete.ModelSelect2(url='contactos:contactos-autocomplete',forward=['cliente_biable ']),
        required=False,
        label='Contacto'
    )
```

At the moment that I took out forward in the last one, it works properly.

In the HTML field I put 

```
{% block contenct %}
     <form action="" method="post">
          {{ form.as_p }}
     </form>
{% endblock %}

{% block footer %}
    <script type="text/javascript" src="{% static 'admin/js/vendor/jquery/jquery.js' %}"></script>
    {{ form.media }}
{% endblock %}
```

Have someone had the same problem? Is there any solution?
Thanks.

jpic:
I don't understand, it should be defined by this script: https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal/static/autocomplete_light/autocomplete.init.js

Perhaps you need to run collectstatic again ? Do you have DEBUG enabled ?

Perhaps this document can help in addition to Django's documentation on static files: http://blog.yourlabs.org/post/30382323418/surviving-djangocontribstaticfiles-or-how-to

Fabitosoft:
Hi,
You got a point there, I have noticed that JQuery has been loaded twice. 
First, when I made the call with media after my form, and after, because it is in the base.html. I have termporally removed it from base.html and it works properly. 
Does someone has any peace of code to resolve this? I mean, to keep the JQuery that I have in base.html but just in this case to change it for the other version? Any good solution to make more clear my script?

Thanks

jpic:
Would it work if you just remove it from your footer block ?

As long as your base.html loads jquery before the footer block it would work.
Otherwise, just move {{ form.media }} after jquery in your base.html ;)

Fabitosoft:
[...]

PR #838, opened 2017-04-24
Use namespaced jQuery in `get_forwards`

### LINKED ISSUE #792: "getFormPrefix is not a function" when using forward.Const
Hi!

I'm using this fantastic library, but then I run into a problem.
Here is a form that I use in a inline tab:

```python
class ComponentImageForm(forms.models.ModelForm):

    class Meta:
        widgets = {
            'image': autocomplete.ModelSelect2(
                url='image_autocomplete',
                attrs={
                    'data-html': 'true'
                }
            )
        }
```

It works as expected.
I need to search only for product images, and this is a hard coded key on my image model.
So, adding forward to the mix:

```python
class ComponentImageForm(forms.models.ModelForm):

    class Meta:
        widgets = {
            'image': autocomplete.ModelSelect2(
                url='image_autocomplete',
                forward=(
                    forward.Const('product', 'image_application'),
                ),
                attrs={
                    'data-html': 'true'
                }
            )
        }
```

That gives me a "getFormPrefix() is not a function" error.

What I'm doing wrong?

Thanks!

ryan-copperleaf:
The function `get_forwards` was defined without the usual IIFE aliasing `yl.jQuery` to `$`. Since it wasn’t used anywhere else, this also moves it inside the Select2 script to avoid defining the function globally.

Should fix #792 (#806).


ryan-copperleaf:
(Linting is failing because of https://gitlab.com/pycqa/flake8-docstrings/merge_requests/7.)

jpic:
I got a good feeling about this, let's merge it and we can still revert / work it out if @gagarski (original author of this script) finds any objection.

gagarski:
It is not used outside select2.js but actually this code could be used for other autocomplete widgets. For now, I think we can move it back and/or refactor it when there will be any.

ryan-copperleaf:
Thanks, everyone!

PR #839, opened 2017-04-25
Add checkbox handling in forwards

marekjedrzejewski:
+ also add simple test that only outputs the checkbox state to console

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -35,16 +35,21 @@
                     dstName = srcName;
                 }
                 // First look for this field in the inline
-                $field = $('[name=' + prefix + srcName + ']');
-                if (!$field.length)
+                $field_selector = '[name=' + prefix + srcName + ']';
+                $field = $($field_selector);
+                if (!$field.length) {
                     // As a fallback, look for it outside the inline
-                    $field = $('[name=' + srcName + ']');
-                if ($field.length)
-                    if($field.attr('type') === 'checkbox')
+                    $field_selector = '[name=' + srcName + ']';
+                    $field = $($field_selector);
+                }
+                if ($field.length) {
+                    if ($field.attr('type') === 'checkbox')
                         forwardedData[dstName] = $field[0].checked;
+                    else if ($field.attr('type') === 'radio')
+                        forwardedData[dstName] = $($field_selector + ":checked").val();
                     else
                         forwardedData[dstName] = $field.val();
-
+                }
             }
         });
         return JSON.stringify(forwardedData);
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -40,7 +40,10 @@
                     // As a fallback, look for it outside the inline
                     $field = $('[name=' + srcName + ']');
                 if ($field.length)
-                    forwardedData[dstName] = $field.val();
+                    if($field.attr('type') === 'checkbox')
+                        forwardedData[dstName] = $field[0].checked;
+                    else
+                        forwardedData[dstName] = $field.val();
 
             }
         });
--- a/test_project/rename_forward/forms.py
+++ b/test_project/rename_forward/forms.py
@@ -6,13 +6,18 @@ from .models import TModel
 
 
 class TForm(forms.ModelForm):
+    checkbox = forms.BooleanField(label='checkbox',
+                                  required=False,
+                                  initial=False)
+
     class Meta:
         model = TModel
-        fields = ('name', 'owner', 'test')
+        fields = ('name', 'checkbox', 'owner', 'test')
         widgets = {
             'test': autocomplete.ModelSelect2(
                 url='linked_data_rf',
-                forward=(forward.Field(src="owner", dst="possessor"),
+                forward=(forward.Field(src="checkbox"),
+                         forward.Field(src="owner", dst="possessor"),
                          forward.Const(val=42, dst="secret"))
             )
         }
--- a/test_project/rename_forward/urls.py
+++ b/test_project/rename_forward/urls.py
@@ -8,7 +8,7 @@ from .models import TModel
 class LinkedDataView(autocomplete.Select2QuerySetView):
     def get_queryset(self):
         qs = super(LinkedDataView, self).get_queryset()
-        print(self.forwarded.get('checkbox', None))
+
         possessor = self.forwarded.get('possessor', None)
         secret = self.forwarded.get('secret', None)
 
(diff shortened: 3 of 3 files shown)


test_project/rename_forward/urls.py
```diff
@@ -8,6 +8,7 @@
 class LinkedDataView(autocomplete.Select2QuerySetView):
     def get_queryset(self):
         qs = super(LinkedDataView, self).get_queryset()
+        print(self.forwarded.get('checkbox', None))
```
jpic: Can we remove that line ?

marekjedrzejewski: Of course, that's purely for testing purposes. I didn't have idea for a meaningful test, but had to test it somehow.

jpic:
I think we should not worry about tests on this because they are under development in #833.

If this patch is good for @gagarski it's good for me, feel free to merge or just approve <3

test_project/rename_forward/urls.py
jpic: Another way of testing that's doesn't require code instrumentalization with print statements is to use a debugger. Such as pdb or ipdb, ever heard of it ? It's pretty cool imho <3

marekjedrzejewski: Yeah, I'm using it for my django project all the time as it's the part of python mode in spacemacs.

I submited this along with this cause I assumed it is better to have some test, than none, even if it would be thrown away in a momend, and pdb/ipdb would stop the execution.

jpic: Yeah true, it's preferable to remove before pushing the code because then we can just merge it when someone feels like it ;)

marekjedrzejewski: Understood. Next time I make a pull request, and tests aren't as they should be, don't push them. :)

gagarski:
I've also added support for radio buttons (if anyone uses them). Unfortunately I didn't find a way to add radio buttons into test project (seems like inline model form does not like them).

src/dal_select2/static/autocomplete_light/select2.js
```diff
@@ -35,13 +35,20 @@
...
-                    $field = $('[name=' + srcName + ']');
-                if ($field.length)
-                    forwardedData[dstName] = $field.val();
-
+                // As a fallback, look for it outside the inline
+                    $field_selector = '[name=' + srcName + ']';
+                $field = $($field_selector);
+                if ($field.length) {
+                    if ($field.attr('type') === 'checkbox')
+                        forwardedData[dstName] = !!$($field_selector + ":checked").val();
```
jpic: I don't know if val() would return a string such as "unchecked" then the forwarded value would be `true` with the `!!` syntax. Did you consider this ?

src/dal_select2/static/autocomplete_light/select2.js
```diff
@@ -35,13 +35,20 @@
...
+                    else if ($field.attr('type') === 'radio')
+                        forwardedData[dstName] = $($field_selector + ":checked").val();
```
jpic: Have you tried [$.serializeArray](http://api.jquery.com/serializeArray/) ? It makes a json object with form data, perhaps this would avoid having to maintain "form to json" logic in this script ?

test_project/rename_forward/forms.py
```diff
@@ -6,13 +6,18 @@
...
     class Meta:
         model = TModel
-        fields = ('name', 'owner', 'test')
+        fields = ('name', 'checkbox', 'owner', 'test')
         widgets = {
             'test': autocomplete.ModelSelect2(
                 url='linked_data_rf',
-                forward=(forward.Field(src="owner", dst="possessor"),
+                forward=(forward.Field(src="checkbox"),
+                         forward.Field(src="owner", dst="possessor"),
```
jpic: Way cool, you got it :1st_place_medal:

src/dal_select2/static/autocomplete_light/select2.js
gagarski: I was trying to be too clever and reuse `$($field_selector + ":checked")` for checkboxes and radiobuttons. But still had separate branches.

Actually it would work. `$($field_selector + ":checked")` would not select any checkboxes and `.val()` would return `undefined` which is casted to `false`. I agree that @marekjedrzejewski's way was more straightforward so I reverted it back.

gagarski:
Regarding $.serializeArray.
I experimented with this method.
`$('[name="checked_checkbox"]').serializeArray()` gives 
```json
[
{
    "name": "checked_checkbox",
    "value": "value_attribute_of_checkbox"
}
]
```
but `$('[name="unchecked_checkbox"]').serializeArray()` gives
```json
[
]
```

So we still need special handling for checkboxes in `get_forwards()`

jpic:
Have you planned to rewrite history to include changes from "oops" commits in the commits they should be in before merging ? That would be cool when we update CHANGELOG

jpic:
Yep I've noticed it in the documentation. I've thought that the user should not be expecting the JSON to not be tampered with anyway, and rely on a default value, but that might cause other issues then ? I like the new implementation anyway it looks reasonable to have.

gagarski:
I've noticed that there is also `CheckboxSelectMultiple` which also uses checkboxes but has completely different semantics from boolean field checkbox. And now I cannot see an approach to properly handle both of these widgets. Ideally `CheckboxInput` should be forwarded as boolean value or as a single optional value of `value` HTML-attribute of the checkbox, and `CheckboxSelectMultiple` should be forwarded as a list of selected values (even if it has only one element).

gagarski:
I've noticed that there is also `CheckboxSelectMultiple` which also uses checkboxes but has completely different semantics from boolean field checkbox. And now I cannot see an approach to properly handle both of these widgets. Ideally `CheckboxInput` should be forwarded as boolean value or as a single optional value of `value` HTML-attribute of the checkbox, and `CheckboxSelectMultiple` should be forwarded as a list of selected values (even if it has only one element).

One way to cope with this issue is to add a parameter to `forward.Field`, something like `forward.Field(src="checkbox_multiple", multiple=True)`. I do not like this idea very much. Moreover some fields (like `select` with `multiple` attribute) should be implicitly forwarded as a list.

gagarski:
> I've noticed that there is also CheckboxSelectMultiple which also uses checkboxes but has completely different semantics from boolean field checkbox.

Perhaps it's not related to the current PR, so we should merge it.

jpic:
Implicit forward as a list makes sense for me too.

gagarski:
@jpic is it OK for you to merge this one and work out forwarding a `CheckboxSelectMultiple` in other branch (probably rewriting `get_forwards()` to use `serializeArray`)?

jpic:
Of course, you can merge too @gagarski you're good if not better than me at what we're doing now <3

PR #844, opened 2017-05-02
Improve 'data-html' conditional in select2.js

jhrr:
With `widget=autocomplete.ModelSelect2(attrs={"data-html": True})` an empty attribute, `data-html`, is set on the relevant html node telling the javascript to render the `get_result_label` text out as html rather than text.

However, when it comes to the test `if (element.attr('data-html') !== undefined)` the value being assessed is the empty string and so that test is **failing** and the markup we want to render is just coming out as a string, tags and all.

This diff improves that conditional so that the test passes on the empty string and the html is correctly rendered out as we would expect.

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -60,7 +60,8 @@
 
         // Templating helper
         function template(item) {
-            if (element.attr('data-html') !== undefined) {
+            var attr = element.attr('data-html');
+            if (typeof attr !== typeof undefined && attr !== false) {
                 var $result = $('<span>');
                 $result.html(item.text);
                 return $result;


gagarski:
As for me looking in this code, the old condition should've worked properly. 

With `widget=autocomplete.ModelSelect2(attrs={"data-html": True})` we have HTML for widget like `<select data-html ... dal-autocomplete-function="select2">...</select>`
`element.attr('data-html')` for this `select` returns `""`. Because `"" !== undefined` is true in JS (check it in the browser console) the `if` branch is executed and HTML is rendered in Select2 items.

With `widget=autocomplete.ModelSelect2(attrs={"data-html": False})` (or no attrs at all) we have HTML like `<select ... dal-autocomplete-function="select2">...</select>`
`element.attr('data-html')` for this `select` returns `undefined`. `undefined !== undefined` is false in JS, we're going to the `else` branch and rendering text in Select2 items.

I rechecked this just now: it works as intended in the following cases:
 - `widget=autocomplete.ModelSelect2(attrs={"data-html": True})` renders items as HTML;
 - `widget=autocomplete.ModelSelect2(attrs={"data-html": False})` renders items as plain text;
 - `widget=autocomplete.ModelSelect2(attrs={"data-html": "true"})` renders items as HTML (old client code is unaffected).


Could you please explain me where am I wrong?  Or provide an example where old code works incorrectly?

jhrr:
Ah, my apologies, it looks like the conditional was updated in commit https://github.com/yourlabs/django-autocomplete-light/commit/32097207e10d0dfff4c3a0c046f997237815e6bf a couple of days ago.  I was working around the previous `(element.attr('data-html'))` which is no longer the case.

Feel free to close this, sorry for the static. Thanks for the effort.

gagarski:
@jhrr yeah, previously you had to write `{"data-html": "true"}` in Python code. It was stated in the docs but writing `"true"` instead of `True` Python code seemed ugly, so I fixed it.

jhrr:
It's better! Thanks very much.

PR #849, opened 2017-05-14
Don't create a new option if an iexact-matching one already exists

liwenyip:
# Current behaviour:

Select2 offers to create a new option even if an identical one exists (and if you click on it, it will actually create a duplicate in your database unless you've set up your models to prevent duplicates)

image

# Desired behaviour:

Select2 doesn't offer to create a new option if an identical one exists

image

Or if a case-insensitive match exists

image

Only offer to create a new option if an exact match doesn't exist

image

# Notes

Probably also need to do something similar in `Select2ListView.get()`

### DIFF
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -33,6 +33,10 @@ class Select2ViewMixin(object):
             if page_obj is None or page_obj.number == 1:
                 display_create_option = True
 
+            # Don't offer to create a new option if a (case-insensitive) identical one already exists
+            if self.get_queryset().filter(**{self.create_field +'__iexact': q}).exists():
+                display_create_option = False
+
         if display_create_option and self.has_add_permission(self.request):
             create_option = [{
                 'id': q,
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -33,8 +33,11 @@ class Select2ViewMixin(object):
             if page_obj is None or page_obj.number == 1:
                 display_create_option = True
 
-            # Don't offer to create a new option if a (case-insensitive) identical one already exists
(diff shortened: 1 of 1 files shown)


jpic:
Nice report ! Is there any way to fix this without making another SQL query ?

liwenyip:
Good point @jpic

Yes, we can avoid another SQL query by doing the check in render_to_response() where the queryset is already in context.

I'll make another commit to this PR (you can squash the commits when you merge, right?)

jpic:
Absolutely, will you be able to add a couple of unit tests too for the cases where it exists and where it doesn't ?

liwenyip:
@jpic Can do

Because I'm still new to test-driven development:

- can you please point me to where I should write the tests? (`/src/dal_select2/test.py`?)
- Is it enough to just test that the autocomplete view behaves as expected, or do I need to test with a form and make sure the form widget behaves as expected?

I should be able to figure out the rest from there.

Thanks in advance

jpic:
<3 the patch !

Tests for py34 and dj18 and dj110 did not pass. I re-launched them manually on travis. If they fail again it is likely that it requires some additionnal patches.

See:

https://travis-ci.org/yourlabs/django-autocomplete-light/builds/233401780?utm_source=github_status&utm_medium=notification

jpic:
src/dal_select2/tests.py will do or if you prefer to start fresh src/dal_select2/test_views.py

You can just do a unit test for each of your case:

- add item in db -> execute view -> assert create option **not** in result
- remove item from db -> execute view -> assert create option in result

jpic:
Your patch passes existing tests !

Gratz for this beautiful patch it's gonna be a piece of art <3

jpic:
Integrated in #954 will be in next release

PR #860, opened 2017-06-05
dal: widgets: use the name if we don't have the id

xrmx:
On a fairly involved configuration where my admin
is using django-mptt-admin and dal_admin_filters and so
django-autocomplete-light it happens that the rendered
widget has no attrs.

```
File "/venv/src/django_mptt_admin/admin.py" in get_queryset
  361. self.filter_specs, self.has_filters, remaining_lookup_params, filters_use_distinct = self.get_filters(request)

File "/venv/src/django/contrib/admin/views/main.py" in get_filters
  114. spec = list_filter(request, lookup_params, self.model, self.model_admin)

File "/venv/src/dal_admin_filters/__init__.py" in __init__
  48.  name=self.parameter_name, value=self.used_parameters.get(self.parameter_name, '')

File "/venv/src/django-autocomplete-light/src/dal/widgets.py" in render
  151. conf = self.render_forward_conf(attrs['id'])

Exception Value: 'NoneType' object has no attribute '__getitem__'
```

To workaround that if we don't have attrs we can use the name instead.

### DIFF
--- a/src/dal/widgets.py
+++ b/src/dal/widgets.py
@@ -147,8 +147,12 @@ class WidgetMixin(object):
 
     def render(self, name, value, attrs=None):
         """Calling Django render together with `render_forward_conf`."""
+        try:
+            field_id = attrs['id']
+        except (KeyError, TypeError):
+            field_id = name
         widget = super(WidgetMixin, self).render(name, value, attrs)
-        conf = self.render_forward_conf(attrs['id'])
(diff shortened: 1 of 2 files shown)


jpic:
Nice ! Do you want to throw in a couple of unit tests too ?

xrmx:
Updated PR wit tests

jpic:
Sorry for the late reply, integrated in #954 will be in next release

PR #864, opened 2017-06-17
Do not push element to initialized set if autocompleteLightInitialize listener is not yet set

apinsard:
When loading scripts asynchronously with requirejs for instance, the `initialize()` method may be called before the `autocompleteLightInitialize` listener is set. This was fixed in #820. However, the element is still pushed to `initialized`, which causes subsequent calls to be pointless.

To fix this, I suggest to set a global variable `__dal__initListenerIsSet` right after the listener is set and prevent the execution of the function until it is set to `true`.


jpic:
Nice ! !!

Paul424:
Antoine,

It's too complex for me to follow. What i did in the end to finally solve .js errors from happening due to race-conditions (while loading these select2/autocomplete .js scripts through ajax calls, i.e. basically replacing elements in the DOM), was to load the required .js scripts in advance (on page load that is) and when the content comes in through ajax, i simple call: $("select[data-autocomplete-light-function]").trigger('autocompleteLightInitialize');

I guess in the end that is an easier approach to prevent these issue's from happening.

Paul

PR #868, opened 2017-06-22
Fix selector used to allow clearing non-required fields.

zsiciarz:
`is('required')` seems to return `false` regardless of whether the field has the `required` attribute or not. This change uses a more appropriate selector to configure select2 as clearable.


jpic:
Ooops, nice finding !! thanks a heap !!

zsiciarz:
Looks like these build failures are unrelated to the change :)

Issue #871, opened 2017-06-29
Localization


Q-bart:
Why there is no any locale at the project? I could help to translate it into several languages.

jpic:
Translate what?

Q-bart:
All the strings marked for translation, for example, https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal_select2/views.py#L39

Clients complain about such strings.

jpic:
Agreed, looking forward to review a pull request !

Q-bart:
Pull request was sent.

jpic:
:guitar:

Q-bart:
One more thing: does select2 internationalization (https://github.com/select2/select2#internationalization-i18n) is used here in django-autocomplete-light? I asked this because those strings, like "Please enter [3] or more characters", are not translated.

luzfcb:
@Q-bart the standard (no standard) used by select2 in translation filenames is the problem, see #746

jpic:
@Q-bart yep @luzfcb has done most of the work I think you can rebase PR #746 and finish it easy <3

jpic:
Please add the pt_BR case since this PR was started by our friend from Brazil haha

PR #872, opened 2017-06-29
Localization #871

Q-bart:
Add localization files for dal_select2


jpic:
Nice one !

Q-bart:
One more thing: does select2 internationalization (https://github.com/select2/select2#internationalization-i18n) is used here in django-autocomplete-light? I asked this because those strings, like "Please enter [3] or more characters", are not translated.

PR #874, opened 2017-07-02
Fix Django 1.11.3 error

ikcam:
Django-addanother + DAL raised a render error because didn't have the `renderer` attribute.


SebCorbin:
I can confirm, it would be nice to merge this

PR #876, opened 2017-07-03
Select2 url mutable

balinabbb:
With the built in select2 functionality, the data-autocomplete-light-url can be read and modified dynamically.

For Example:
```javascript
var companySelect = $("select[name=company]");
var deliveryAddressSelect = $("select[name=delivery_address]");
companySelect.change(function (e) {
    var url = "/addresses/auto-complete?company_id="
        + e.target.options[e.target.selectedIndex].value;
    deliveryAddressSelect.attr("data-autocomplete-light-url", url);
});
```
If the company changes in the form, I update the query set of the addresses with the help of the query string. Like this I only get the delivery addresses possible for the given company.

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -121,7 +121,9 @@
             var select = $(this);
 
             $.ajax({
-                url: $(this).attr('data-autocomplete-light-url'),
+                url: function () {
+                    return element.attr('data-autocomplete-light-url');
+                },
                 type: 'POST',
                 dataType: 'json',
                 data: {


jpic:
Could we perhaps add the code example that's in your PR description into your commit ?

balinabbb:
Sure thing.

jpic:
K, please ping me when you update the commit ;)

git commit --amend
git push -f

jpic:
A little example in documentation would be really kewl to share some of your insight :)

jodal:
Should this PR be reopened since the change was reverted in commit 43b3d88d5adeaf7088da22cfed5fc120219aab05?

PR #877, opened 2017-07-03
Return proper content-type for reply

mpasternak:
When running behind nginx, when the content-type is not specified, I get the JSON response nicely wrapped inside <body> tag, which will of course give an error on the web ui. 

I'm not sure how to test for this, except the obvious (checking for content-type in all the subclasses). I also wonder, if this won't break anything for someone else. Well. 

Let me know what do you think.


jpic:
Sure, I don't think it's going to break anything, go ahead B)

Issue #879, opened 2017-07-05
Creating an object using create_field doesn't obey overriden get_result_label


cgada:
I don't think this is intended behaviour, though I'd like confirmation.

[This line](https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal/views.py#L119) calls six.text_type on the newly created object, returning text for this newly created object that might be inconsistent if a user has overriden get_result_label().

jpic:
Agreed, looking forward to review a PR !

timdiels:
Actually, I believe it should use `get_selected_result_label` ([downstream issue](https://gitlab.psb.ugent.be/beg-orcae/orcae/issues/733)). For example, I show `name - description` in the list with `get_result_label` and `name` in the badges with `get_selected_result_label`. The text returned in the created option is used for a badge, so I end up with a badge `name - description`.

PR #881, opened 2017-07-06
Replacing call to six with call to predefined get_result_label function

cgada:
Addresses #879.

If a user has overridden get_result_label, a newly created object still uses the six version of the text instead of using the overridden function.


jpic:
Thanks for this great contribution ! Would you like to take the chance to contribute a regression test in the patch ?

PR #826, opened 2017-04-03
Prevent rendering of empty option on multi select

beruic:
Fixing issue #825

Issue #825, opened 2017-03-31
Select2Multiple adds empty choice


beruic:
Whenever I try to use `Select2Multiple` It adds the empty choice to my choices, which doesn't make sense because it is a multi select.

beruic:
Seems like the issue is that the widget adds the empty choice if `self.placeholder` is set. This is done in `WidgetMixin.render_options`. The lines
```
        else:
            if self.placeholder:
                self.choices.insert(0, (None, ""))
```
seems to do it. I guess  the test could be expanded to test for `self.multi =0 False` too.

jpic:
Is this the else for "if self.url" ?

I suppose you're doing a list autocomplete ? I wonder, more details about
your situation could be interresting.

What's self.multi ?

beruic:
Sorry, I forgot a file reference. It's in `src/dal/widgets.py`. Here is a direct link: https://github.com/yourlabs/django-autocomplete-light/blob/22a107df46860d666ff5067e1465c3f67363cc5b/src/dal/widgets.py#L120.

beruic:
Oh, and I am using a `Select2Multiple`, and `self.multi` does not seem to be defined. Did you mean `self.allow_multiple_selected`? That one is `True`.

The empty option is sometimes selected somehow, and really gets in the way for my form treatment.

jpic:
I totally agree, I'm trying to figure out why this hack is there in the first place, and what does it break if we remove it.

beruic:
Cool. Otherwise, I think my PR will fix it.

jpic:
I'm really not sure, because usually tests pass on django  up to 1.10 and in your PR there is no django version that passes tests ...

beruic:
Oh? I admit that I just fixed it in the browser, but I cannot see how this breaks a lot of things. Perhaps the tests themselves are wrong?

jpic:
The tests are mostly (90%) automated selenium tests for the test apps in the django admin.

jpic:
The tests are mostly (90%) automated selenium tests for the test apps in the django admin, so they don't really care about changes at this level.

beruic:
It doesn't seem like it is my change that made it fail. If you look at https://travis-ci.org/yourlabs/django-autocomplete-light/builds, master seems to have the same issues.

jpic:
This sounds like good news, sorry your pr will have to wait until the tests are fixed ... i wish i had more time to maintain this code ^^

[...]

PR #885, opened 2017-07-11
Prevent rendering of empty option on multi select

johandc:
Fixes: Select2Multiple adds empty choice #825 for Django >1.10
See also PR #826


jpic:
Nice ! Should we have a test case for this ?

beruic:
Probably, but we (@johandc and I) are on vacation at the moment :)

johandc:
This fix is exactly similar to the previous one that has been in the main-line for quite some time.

Did you make a test case back when you fixed this issue for Django <1.10?

With that taken into consideration i belive it is safe to pull this.

jpic:
Thanks, this will be in next release, #953

PR #887, opened 2017-07-14
simple model autocomplete - initial draft

toudi:
This is an initial draft for what I call 'simple autocomplete for models'.

@jpic I would like to deeply and sincerely apologise for this insane delay. Could you please take a look at this initial draft and guide me how it should be secured / enhanced ?

Dear friend developer @EvaSDK from PR #815 - could you take a look at this as well ?

Here's a sample usage:

```
#urls.py
from django.conf.urls import url
from dal_select2.views import Select2GenericModelQuerySetView
from models import Parent # this is my model class, it contains 2 fields - first_name, last_name

urlpatterns = [
    url(r'^$', views.index),
    url(
        r'^ac',
        Select2GenericModelQuerySetView.as_view(
            model=Parent,
            search_fields=('first_name', 'last_name'),
        ),
        name='parent-ac',
    )
]

#views.py
from django.shortcuts import render
from forms import Search

# Create your views here.
def index(req):
    form = Search()

    return render(
        req,
        'foo/test.html', {
            'form': form
        }
    )

#forms.py
from django import forms
from models import Parent
from dal_select2.widgets import ModelSelect2


class Search(forms.Form):
    parent = forms.ModelChoiceField(
        queryset=Parent.objects.all(),
        widget=ModelSelect2(url='parent-ac')
    )

    class Meta:
        exclude = []

```

p.s. I tried to implement this in v2 branch as well, but I failed to see where are the views for select2 :/

### DIFF
--- a/src/dal/views.py
+++ b/src/dal/views.py
@@ -8,6 +8,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.http import HttpResponseBadRequest, HttpResponseNotAllowed
 from django.utils import six
 from django.views.generic.list import BaseListView
+from django.db.models import Q
 
 
 class ViewMixin(object):
@@ -118,3 +119,31 @@ class BaseQuerySetView(ViewMixin, BaseListView):
             'id': result.pk,
             'text': six.text_type(result),
         })
+
+
+class GenericModelQuerySetView(BaseQuerySetView):
+    model = None
+    search_fields = []
+
+    def get_queryset(self):
+        qs = super(BaseQuerySetView, self).get_queryset()
+
+        if self.q and self.search_fields:
+            # this may seem like an awkward code, because one could also write
+            # this as:
+            # _filters = Q()
+            # for ..
+            #   _filters |= Q( .. )
+            # however, this would yield slightly awkward query as a result:
+            # (OR: (AND: ), [ here goes the relevant part ] )
+            _filters = None
+            for field in self.search_fields:
+                _filter = Q(**{'%s__icontains' % field: self.q})
+                if not _filters:
+                    _filters = _filter
+                else:
+                    _filters |= _filter
+
+            qs = qs.filter(_filters)
+
+        return qs
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -4,7 +4,7 @@ import collections
 import json
 import six
 
-from dal.views import BaseQuerySetView, ViewMixin
+from dal.views import BaseQuerySetView, ViewMixin, GenericModelQuerySetView
 
 from django import http
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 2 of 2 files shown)


jpic:
Don't worry about the v2 branch.

I'm all for such an improvement, however, isn't there some library already encapsulating such logic we could depend on ? 

Ie. https://github.com/gregplaysguitar/django-simple-search

Otherwise, shouldn't we make such a package ?

I mean that I'd like to reuse that in a search form or something, I think it makes sense to have this feature with dal, but I'm not convinced that the actual code belongs to dal itself. 

What do you think ?

toudi:
Hmm I'm not so sure whether it would be justifiable to depend on 3rd party library. If you look at the changes I made, you'd see that the changes minus the documentation is something like 15 lines of code. Actually, my code is actually a very thin inheritance of the View already provided by dal - it only makes advantage of different fields than `name`.

I would also like to point out that it is not my goal to extend and extend this class even more, because it would most certainly create the complex config structure in `urls.py`. In my opinion, if one wants something other than this pure simple autocomplete, based on specified fields, this is a good place to start implementing custom views with get_queryset

I have no idea whether django-simple-search provides this feature - I kind of always used dal, and my only papercut was that I had to repeat and repeat these views with get_queryset simply because I wanted to autocomplete by different fields

About the reusing in search form - I don't think I  understand your question. The example which I have added shows the Search form, right ? Or are you referring to some generic Search Form ?

jpic:
Yes of course this is simple enough but in v2 this grew in a pretty empiric way.

So perhaps you might prefer to just port that code in v3 ? It's got a lot of good features:

https://github.com/yourlabs/django-autocomplete-light/blob/v2/autocomplete_light/autocomplete/model.py

toudi:
of course I can try to :D I just hope that I won't be delayed as much as I was before creating this PR.. otherwise by the time I came up with a PR there will already be v4 of dal :D

toudi:
actually, wait a second.. I can see in my comment that I failed to provide a PR for v2 branch, therefore this change should already be compatible with v3.. or ?

jpic:
Yeah sorry about that, you were already on v3 yes ;)

v2 was bloat over v1, v3 is v2 without the bloat ;)

toudi:
hmm so I don't quite understand your comment. Does it mean that this PR is no longer necesary because current v3 already provides such features?

jpic:
v3 doesn't provide such feature, i thought someone would extract that from v2 and release it in a separate app, because you know, it can be useful for any generic list view, not specifically to DAL ;)

about the pr yes coverage would be welcome to help with future python or django version upgrades, coverage drastically reduce maintenance cost over years ;)

also, perhaps we could already use the same search_fields as in v2 ? those were user contributed, they are going to contribute it again anyway if we don't port it i can see them coming haha

jpic:
Do you we should use SearchForm from django-betterforms ?

https://github.com/jpic/django-betterforms/blob/master/betterforms/changelist.py#L66

Then, we could contribute search_fields improvements there.

jpic:
My last comment turned completely wrong as django-betterforms has died since, but 053554937475e1a34a5e48de5bf6d66c90e9de45 pastes the search_fields feature from django.contrib.admin and no one has been complaining so far :joy:

:tophat:

PR #889, opened 2017-07-19
Fix v2 compatibility with django 1.11

lezeroq:
For those who still uses v2 this patch fixes compatibility with django 1.11


jpic:
Thanks for contributing <3

lezeroq:
Sorry, I've made mistake in this PR, so one new PR https://github.com/yourlabs/django-autocomplete-light/pull/896

Issue #845, opened 2017-05-03
Separate custom markup for results and final value


jhrr:
I have a Python autocomplete view returning custom markup. (I need to disambiguate possibly duplicate names using email addresses so I need the name and the email in the search results.)

```python
class MyAutocompleteView(autocomplete.Select2QuerySetView):
    def get_result_label(self, item):
        return f"""
            <div class="modelchoice-search-result">
                <div class="modelchoice-search-name">{item.name}</div>
                <div class="modelchoice-search-email">{item.email}</div>
            </div>
        """
```        

This ultimately generates the html below inside `<span class="select2 select2-container>` (I've omitted a few containers/levels for clarity).

```html
<span class="select2-selection__rendered" id="select2-id_form-13-child-container" title="
    <div class=&quot;modelchoice-search-result&quot;>
        <div class=&quot;modelchoice-search-name&quot;>Firstname Lastname</div>
        <div class=&quot;modelchoice-search-email&quot;>email@example.com</div>
    </div>
"><span class="select2-selection__clear">×</span><span>
    <div class="modelchoice-search-result">
        <div class="modelchoice-search-name">Firstname Lastname</div>
        <div class="modelchoice-search-email">email@example.com</div>
    </div>
</span></span>
```

So, it looks like the markup template is used for both the display of the results and also as the value that is added to the widget upon selection. That's fine, I guess that makes sense as much as not. But is there any way I can separate out the display of the results from the final value that is added to the field? In this case, for example, I want to omit the email and be left just with the name.

I've tried just deleting that DOM node and other hacks, I've tried those by attaching to both the `change` event and also select2's `select2:close` event, etc. But there seems to be some DOM manipulation happening that I'm not able to override and so I'm always ultimately left with the full template being added to the box. Any suggestions would be much appreciated, I'd prefer not to have to fork and hack the template logic in [`autocomplete_light/select2.js`](https://github.com/yourlabs/django-autocomplete-light/blob/32097207e10d0dfff4c3a0c046f997237815e6bf/src/dal_select2/static/autocomplete_light/select2.js#L62).

jpic:
I think you will have better luck with jquery-autocomplete-light if select2's templating system doesn't work for you, PoC in #749 might help

jhrr:
Thanks I'll check it out then.

gagarski:
[...]

PR #892, opened 2017-07-22
Enable different item label for selected item

maximpetrov:
This pull request adds ability to display selected result (in select field) in different way than in dropdown e.g. display detail item in dropdown but only name in field after selection.

### DIFF
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -21,6 +21,7 @@ class Select2ViewMixin(object):
             {
                 'id': self.get_result_value(result),
                 'text': self.get_result_label(result),
+                'selected_text': self.get_selected_result_label(result),
             } for result in context['object_list']
         ]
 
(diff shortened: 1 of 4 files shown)


gagarski:
Should fix #845

jpic:
Looks good for me !

We just need some tests in this PR.

src/dal_select2/views.py
```diff
@@ -21,6 +21,7 @@ def get_results(self, context):
             {
                 'id': self.get_result_value(result),
                 'text': self.get_result_label(result),
+                'selected_text': self.get_selected_result_label(result),
```
mik-laj: I think it is worth adding this field only if it is different from ``text``. Now one text is sent in two fields even when it is identical.

jpic: That's true but we're going to merge this as is so if you got hard feelings about this then perhaps make another pr or open a ticket :joy:

jpic:
Will be in next release.

PR #904, opened 2017-08-28
Fix KeyError when id is not in attrs

dwheaton:
Ran into this issue when adding a User autocomplete to an admin ActionForm. My other autocompletes work fine and this one change allowed this one to work as well. Unsure why this works but perhaps because my form is not a ModelForm?

### DIFF
--- a/src/dal/widgets.py
+++ b/src/dal/widgets.py
@@ -148,7 +148,7 @@ class WidgetMixin(object):
     def render(self, name, value, attrs=None):
         """Calling Django render together with `render_forward_conf`."""
         widget = super(WidgetMixin, self).render(name, value, attrs)
-        conf = self.render_forward_conf(attrs['id'])
+        conf = self.render_forward_conf(attrs['id']) if 'id' in attrs else ""
         return mark_safe(widget + conf)
 
     def _get_url(self):


jpic:
Your patch disables the field forward feature. I thought all django form elements would have an id auto generated ?

dwheaton:
I thought so too... I tried manually adding the id attr but didn't fix my
issue.

I think the patch shouldn't break anything since it only avoids a KeyError
exception if/when id isn't in attrs.

But I wasn't able to figure everything out from my debugging and I don't
use the field forward feature so whatever you think!

Thanks,

--David


On Aug 30, 2017 1:26 AM, "jpic" <notifications@github.com> wrote:

Your patch disables the field forward feature. I thought all django form
elements would have an id auto generated ?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/yourlabs/django-autocomplete-light/pull/904#issuecomment-325920131>,
or mute the thread

jpic:
I haven't yet tested your patch, but it seems that conf will be empty if the id attr is not set here:

jpic:
I haven't yet tested your patch, but I think this code will make conf=='' if there is no id attribute:

    conf = self.render_forward_conf(attrs['id']) if 'id' in attrs else ""

You can use your fork which works while we find a better solution perhaps?

dwheaton:
Yes that's the effect but only if id isn't set...and if that is the case
then then an exception is raised.

I can get by for now. Thanks for testing and considering it when you have
time.

jpic:
If you want to help with this then creating an example app in the test_project to reproduce would be great. We could then automate a test on it.

dwheaton:
Okay sure will try to reproduce properly when I'll have time.

jpic:
Thanks, will be in next release, #953

PR #916, opened 2017-09-26
Fix a mistake in the documentation

voidus:
I'm not 100% certain about this, so this PR is also a request for clarification.


jpic:
You are absolutely correct !

Thanks !

Issue #918, opened 2017-09-30
Allow every registered user to create objects 


davideghz:
`Note that creating objects is only allowed to staff users with add permission by default.`

It would be great to pass an option from `urls.py` like
```
views.MyModelAutocomplete.as_view(create_field='name', public=True)
```
to skip the permission check here: http://django-autocomplete-light.readthedocs.io/en/master/_modules/dal/views.html#BaseQuerySetView.post

I would love to contribute with a PR but I still don't feel that confortable with Python/Django :)

In the meantime, is there any way to make this option available for all the subscribed users? Thanks!

jpic:
What happens after you grant the create permission for that model to your registered user ? It seems like that's what [has_add_permission](http://django-autocomplete-light.readthedocs.io/en/master/_modules/dal/views.html#BaseQuerySetView.has_add_permission) is doing.
Perhaps we we have misleading documentation somewhere ?

davideghz:
Yeah it actually works and I believe it's a way safer way to handle the point. I was approaching the problem from the other side.

The misleading part in the docs is the reference to `only allowed to staff users`. I was induced to think that only `staff users` could create objects, but I can actually grant `add` permission to a user even if it's no staff.

I would change it to `Note that creating objects is only allowed to staff users with add permission by default. If you want to grant add permission to other users you have to explicitly set it.` and then add a noob-proof-example, like:
```
permission = Permission.objects.get(name='Can add model-name')
user.user_permissions.add(permission)
```

jpic:
Can you confirm that it works with any authenticated user which has create permission and that is_staff is not required anymore ?
Then we should update the documentation yes

davideghz:
Yes, I do confirm that `is_staff` is no longer required. I can contribute to the docs if you explain me how :)

jpic:
Sure, just make a pull request with changes in the docs/ subdirectory.

davideghz:
Open source noob here :) what's the correct flow? I cloned the repo but then I'm not allowed to push any new branch to origin..

davideghz:
oh, got it https://github.com/yourlabs/django-autocomplete-light/pull/919

PR #919, opened 2017-10-01
Improve create-new-object docs with example

davideghz:
https://github.com/yourlabs/django-autocomplete-light/issues/918#issuecomment-333366515

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -348,9 +348,8 @@ This way, the option 'Create "Tibet"' will be available if a user inputs
 the view which will do ``Country.objects.create(name='Tibet')``. It will be
 included in the server response so that the script can add it to the widget.
 
-Note that creating objects is only allowed to staff users with add permission
-by default. If you want to grant add permission to other users you have to
-explicitly set it.
+Note that creating objects is allowed to staff users by default. If you want
+to grant add permission to other users (not staff users) you have to explicitly set it.
 
 .. code-block:: python
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -348,8 +348,9 @@ This way, the option 'Create "Tibet"' will be available if a user inputs
 the view which will do ``Country.objects.create(name='Tibet')``. It will be
 included in the server response so that the script can add it to the widget.
 
-Note that creating objects is allowed to staff users by default. If you want
-to grant add permission to other users (not staff users) you have to explicitly set it.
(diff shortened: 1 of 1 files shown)


davideghz:
umh, why does it break tests?

docs/tutorial.rst
```diff
@@ -349,7 +349,13 @@ the view which will do ``Country.objects.create(name='Tibet')``. It will be
 included in the server response so that the script can add it to the widget.
 
 Note that creating objects is only allowed to staff users with add permission
```
jpic: Should we change this so that we don't need to be staff user anymore ?

jpic:
Don't worry about it, some changes occured upstreams I'll fix them when adding dj2.0 ;)

Do we really still need to be staff user ?

docs/tutorial.rst
davideghz: The point is. Staff members get this privilege by default; to grant same right to other users (not staff members) it's necessary to set the `add` permission explicitly.

jpic:
I'm really not sure, I don't see any requirement to have a user with is_staff=True in the source code, are you sure this is necessary ?

davideghz:
I see what you mean. As per my tests, a staff user was able to create new objects, while a standard user wasn't. As per Django doc, however, being a staff user does not implies having create rights on resources. I will try it out on a new test project and follow up!

jpic:
I've just tested in the test project and I can confirm that is_staff is not necessary anymore, only being logged in and having create permission is required.

Could you update the docs to match this please ?
Thanks a HEAP for your effort, rock on :guitar: :saxophone: :trumpet:

davideghz:
Cool! Thx for testing!

jpic:
Just perfect, your example really helped me out !
But at the end of the day I was using Permssion.objects.get(name='add_mymodel') which was a bit easier than filtering by name, did you try that yourself ?

davideghz:
well I use it in a `post_save` signal exactly as I wrote and it works perfectly :D
My model is called `DjProfile` and I use it like:
```
@receiver(post_save, sender=User)
def grant_permissions(sender, instance, created, **kwargs):
    if created:
        permission = Permission.objects.get(name='Can add dj profile')
        instance.user_permissions.add(permission)
```
And yeah, maybe using groups would even be a better approach than setting the permission user by user.. I will think about it :)

jpic:
Nice, thanks a heap for your contribution, be sure that it will help the community ;)

Best <3

PR #920, opened 2017-10-02
Dj2 & test fix

jpic:


jpic:
Rewrote history, follow up on #933

PR #921, opened 2017-10-02
Fix Incompatibility with django-nested-admin

### LINKED ISSUE #917: Incompatible with django-nested-admin
This package works wonder (thanks) by itself but doesn't forward the forward value correctly when used with [django-nested-admin](https://github.com/theatlantic/django-nested-admin/).

This has to do with the `get_forwards` which falls back [only one level](https://github.com/yourlabs/django-autocomplete-light/blob/08d69f0938836d511e1866f77286237c6e4f897b/src/dal_select2/static/autocomplete_light/select2.js#L37-L44).

I was able to get it working by first creating a `getFormPrefixes` which will return all the possible prefixes instead of just the most immediate one like `getFormPrefix`.

```javascript
/*
 * Returns a list of prefixes from the most specific to the least.
 * This is to work with multiple nested levels.
 */
$.fn.getFormPrefixes = function() {
    var parts = $(this).attr('name').split('-').slice(0, -1);
    var prefixes = [];

    for (i = 0; i < parts.length; i += 2) {
        var testPrefix = parts.slice(0, -i || parts.length).join('-');
        if (!testPrefix.length)
            continue;

        testPrefix += '-';

        var result = $(':input[name^=' + testPrefix + ']')

        if (result.length)
            prefixes.push(testPrefix);
    }

    prefixes.push('');

    return prefixes;
}
```

And then, replacing `prefix = $(element).getFormPrefix();` by `prefixes = $(element).getFormPrefixes();` and replacing the [fallback mechanism](https://github.com/yourlabs/django-autocomplete-light/blob/08d69f0938836d511e1866f77286237c6e4f897b/src/dal_select2/static/autocomplete_light/select2.js#L37-L44) by a for loop which breaks on the first match:

```javascript
for (var i = 0; i < prefixes.length; i++) {
    $field_selector = '[name=' + prefixes[i] + srcName + ']';
    $field = $($field_selector);

    if ($field.length) {
        // Same logic for checking for checkbox, radio or else here.

        break;  // break after first match
    }
}
```

Note that in addition of the issues below, #909 and theatlantic/django-nested-admin#84 will need to be fixed too in order to achieve full compatibility.

loicteixeira:
Instead of only falling back one level, check all possible form prefixes, from the most specific to the least and break on first match.

Fix #917.

As I understand it, `test_project.linked_data.test_functional` and `test_project.rename_forward.test_functional` already tests that `forward` values works so I did not add additional tests for that.
If you wanted to have test with `django-nested-admin` just let me know. I just wasn't sure if you want to commit to it or not.


jpic:
Thanks, we need tests, example app and doc update.

loicteixeira:
Currently the tests fail with some random errors (e.g. `InterpreterNotFound: pypy` or `ERROR:   checkqa: commands failed`) which are clearly unrelated to the changes. I branched off `master` but maybe this is not what I was supposed to do. Any pointers?

jpic:
Use one of the environments that work, or start from the dj2 branch, where all tests pass - except with django 2 :D

jpic:
Thanks ! This will be included in next release. #953

Nice work B)

PR #927, opened 2017-10-15
Fix Select2 initializer calls other select initializers too early

### LINKED ISSUE #909: Select2 initializer calls other select initializers too early
If you load the JS file synchronously at the end of your body, and have a custom `autocomplete-light-function`, this initializer might not be called if you also include the Select2 initializer. This is caused by the fix for #820. The extra fix in #864 does not prevent this issue.

### My scenario:

```html
<html>
  <body>
    <select data-autocomplete-light-function="my-custom-initializer" data-autocomplete-light-url="/autocomplete/data" id="id_myfield" name="myfield" data-html required>
      <option value="" selected="selected">---------</option>
    </select>
    
    <script type="text/javascript" src="/static/autocomplete_light/jquery.init.js"></script>
    <script type="text/javascript" src="/static/autocomplete_light/autocomplete.init.js"></script>
    <script type="text/javascript" src="/static/autocomplete_light/vendor/select2/dist/js/select2.full.js"></script>
    <script type="text/javascript" src="/static/autocomplete_light/select2.js"></script>
    <script type="text/javascript" src="/static/my-app/js/my-autocomplete.js"></script>
  </body>
</html>
```

By the time the code in `select2.js` gets executed, `my-autocomplete.js` has not loaded yet. So when it executes this block of code:

```javascript
$('[data-autocomplete-light-function]:not([id*="__prefix__"])').each(function() {
    window.__dal__initialize(this);
});
```

It will find the select box (because the JS is loaded at the end), and it will try to initialize it. The `initialize` (or `__dal__initialize`) method will mark it as initialized, even if the correct `autocomplete-light-function` to execute does not exist yet. So calling this block again will not work. The fix from #864 also does not work in this case, because `__dal__initListenerIsSet` is set by the `select2.js` script, but that is not the initializer I need!

### Possible fix in `select2.js`
Either:
 - Wrap the code in `$(document).ready()`
 - Do not just query for `[data-autocomplete-light-function]`, but for `[data-autocomplete-light-function=select2]`

I haven't tried this out yet (did not have time to look into running the test suite). It might be even better to just rewrite the initialization so the actual method that does the initialization can mark the element as initialized. I'm not sure what the cleanest way to do this is.

### How I got here
I got into this situation because I have a custom widget that extends `ModelSelect2` (mainly custom JS), and it only added to the existing `Media`.

### Workarounds
My current workaround is to copy the whole `Media` block of the `Select2WidgetMixin` superclass and add `extend = False`. Another workaround would be to include the JS at the top of the page (then the block of code will not have an effect because the jQuery selector will not match anything yet).

loicteixeira:
This implements what has been discussed in the issue.

Fix #909.


jpic:
Seems like you nailed it can you rebase on dj2 branch ?

loicteixeira:
Done, but because the PR is based off master, it now shows some unrelated changes and commits.

jpic:
Integrated this in #953 will be in next release.

Thanks !

jpic:
Thanks ! This will be included in next release. #953 

Nice work B)

PR #932, opened 2017-10-29
 Fix #930: QuerySetSequence querysets order is not preserved

### LINKED ISSUE #930: QuerySetSequence querysets order is not preserved
I think I traced the problem to here:

https://github.com/yourlabs/django-autocomplete-light/blob/3.2.10/src/dal_select2_queryset_sequence/views.py#L33

An ordinary dict is being used for grouping the results, and the response after that becomes unordered.

melvyn-sopacua:
Use an ordered dict to preserve ordering of querysets. Since the test
suite isn't working for the most part (changed css in admin) and some
tests, do not seem to be  called at all by tox, we don't add a test for
it.
In particular, `dal_queryset_sequence/tests/test_views.py` isn't being
called as far as I can tell (used --results-json).

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -235,9 +235,12 @@ to return HTML code.
 
 .. code-block:: python
 
+    from django.utils.html import format_html
+    
     class CountryAutocomplete(autocomplete.Select2QuerySetView):
         def get_result_label(self, item):
-            return '<img src="flags/%s.png"> %s' % (item.name, item.name)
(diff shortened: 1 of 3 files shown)


jpic:
What happens if you try `git rebase -i upstream/master` and only keep the lines with your own commits ? Does it update your branch correctly ?

melvyn-sopacua:
Separating into branch to not have my origin/master tainted.

PR #937, opened 2017-11-16
Update tutorial.rst to fix XSS in the example

hangtwenty:
Hello,

I appreciate the existing warning about XSS, but since many will just copy and paste the example given here, I think it is crucial to make the example safe from XSS. Simply using `format_html` does the trick.

(Note: I have noticed that folks sometimes think `%` operator could be overloaded to do escaping automagically - so they might not even realize that what they are copy-pasting is not safe yet.)

Thanks for your consideration on this matter!

Best
Michael

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -235,9 +235,12 @@ to return HTML code.
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


jpic:
Thanks !

PR #933, opened 2017-10-29
Django 2.0 Support

jpic:
Because, sports & love

Issue #923, opened 2017-10-08
Python 3.5 - is it causing my problems?


marteloge:
Using Python 3.5 and is having some import issues.

INSTALLED_APPS = [
    'dal',
    'dal_select2',
    'taggit',
    'autocomplete_light',
    'django.contrib.admin',
    ....
]
ImportError: No module named 'autocomplete_light'

Whe trying to install "python3 -m pip install django-autocomplete-light" I get:
Requirement already satisfied: django-autocomplete-light in /Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages

Is this a python error or maybe something else I am doing wrong?

marteloge:
User error :) Used old documentation

PR #944, opened 2017-12-06
Fix #943: user.is_authenticated is a property

### LINKED ISSUE #943: BaseQuerysetView: user.is_authenticated Django 2.0 compatibility
DAL breaks when rendering the API views because of a call to `request.user.is_authenticated()`, which is deprecated as a method in Django 2.0. This can be fixed by removing the actual function call and just using the read-only property. This change will remain backwards compatible as suggested by [this](https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.is_authenticated).

melvyn-sopacua:
While in here:
- Add Django 2.0 to test matrix
- Ignore browser logs

Note: upstream packages prevent tests from running correctly (gm2m and generic-m2m).

close: #943

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ docs/build
 docs/source/_static
 test_project/htmlcov/
 test_project/.coverage
+test_project/*.log
 build
 dist
 django_autocomplete_light.egg-info/
--- a/src/dal/views.py
+++ b/src/dal/views.py
@@ -92,7 +92,7 @@ class BaseQuerySetView(ViewMixin, BaseListView):
 
     def has_add_permission(self, request):
         """Return True if the user has the permission to add a model."""
-        if not request.user.is_authenticated():
+        if not request.user.is_authenticated:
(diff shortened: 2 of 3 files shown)


melvyn-sopacua:
I see this is also done in #933 - are you going ahead with that one James?

Tests are broken now cause of upstream :( I've forked gm2m for a fix.

jpic:
We should go with #933 because it also fixes a lot of checks that were broken by merges.

jpic:
Hi Melvyn !

Tests are not passing on travis, but they work fine on various versions for me with any browser (phantomjs, firefox and chrome).

However, we don't have the gm2m part ported, dependencies that weren't dj20 compatible at the time were left out from the test runner.

Do you have it working ?

Do you want to rebase your work so that we include it for the next release ?

Keep up the great work :guitar:

jpic:
Taken care of in #923, thanks !

PR #961, opened 2018-01-12
Documentation: Adjusting virtualenv to use Python 3

mogoh:
Django 2.0 doesn't support python 2 any more, thus we need to be sure to use python 3 in the virtual env.


jpic:
Thanks !

PR #970, opened 2018-02-12
A problem with a selected item's id.

Chipelin:
If a selected item's id equal zero  ModelSelect2 widget returns as an empty field.

### DIFF
--- a/src/dal/widgets.py
+++ b/src/dal/widgets.py
@@ -112,7 +112,7 @@ class WidgetMixin(object):
 
         # Filter out None values, not needed for autocomplete
         selected_choices = [six.text_type(c) for c
-                            in args[selected_choices_arg] if c]
+                            in args[selected_choices_arg] if c is not None] # FIXED: c=0 doesn't return a value of None
 
         all_choices = copy.copy(self.choices)
         if self.url:


jpic:
why close

Chipelin:
In my project I use an older version of the autocomplete. 3.2.10 to be exact. Master  has been updated to a higher version already. Didn't make sense to make a pull request if I use an older version. Sorry I'm new to git should've explained before closing.

jpic:
Why not upgrade your DAL version ?

Chipelin:
The latest version doesn't seem to work with Django 1.10 wich I use for my project.

jpic:
Perhaps you can upgrade to 1.11 then it will work ;)

Chipelin:
Perhaps, or perhaps something else will break ;) I've made a quick fix for now on my fork I'll try updating to 1.11 later.

jpic:
k, i've found some simple compatibility patch we can have in current master for older version support, will keep you updated here ;)

Chipelin:
Ok. Thank you  ;)

jpic:
Can you try with the backward_compat branch and see if that works for you ?

jpic:
You can with pip uninstall django-autocomplete-light; pip install -e git+https://github.com/yourlabs/django-autocomplete-light.git@backward_compat#egg=django-autocomplete-light

Chipelin:
I'll try as soon as I get to work (wich will be tomorrow morning in my time zone) and report back.

Chipelin:
So... the project compiled with the version you provided, but my original problem is still present. If I select an element with an id of 0 from the dal widget then save the form, the widget returns empty when I load it. The data is saved correctly to my database. If  I turn the dal widget off the field displays correctly. The problem is present only if element's id = 0. Changing the id is not an option since the table is used in multiple projects.

My fix still works on the new version, however, I realize, that it is probably in no way comprehensive. Should I reopen my pull request?

jpic:
In this case I propose to merge the backward_compat branch into master, merge your PR into master, and make a new release, do you think that would work ? Could you perhaps add a test with id=0 with your patch to ensure it won't break in the future ?

Chipelin:
I think that would work just fine (I mean, it works on my end at least), but I would need to figure out how to write tests properly first ;) It might take some time, considering I have some work I need to do, but I will report back as soon as I can.

PR #976, opened 2018-02-14
Use http.JsonResponse when the response is JSON

mpasternak:
**Warning** this is untested. I have splinter errors when running tox.

http.HttpResponse with manual json.dumps call is, well, kind of acceptable. Unless you forget to set content_type to „application/json”. But then, everything still is fine, tests pass and stuff. But then, after some time, you run the code in the production, on, say, nginx+uwsgi. This combo will attach HTML tags to your json responses, head, body, the html tag itself, thus rendering json responses as unreadable to the client.

TL;DR: Use http.JsonResponse when the response is JSON

### DIFF
--- a/src/dal_select2/views.py
+++ b/src/dal_select2/views.py
@@ -55,15 +55,13 @@ class Select2ViewMixin(object):
 
         create_option = self.get_create_option(context, q)
 
-        return http.HttpResponse(
-            json.dumps({
+        return http.JsonResponse(
+            {
                 'results': self.get_results(context) + create_option,
                 'pagination': {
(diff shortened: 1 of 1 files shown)


mpasternak:
(I actually created this PR in hope that TravisCI will handle the testing for me xfingersx xcrossedx)

jpic:
Works for me ;)

Go ahead and merge when you're ready

mpasternak:
@jpic do you want me to merge this myself, did I understand correctly?

jpic:
@mpasternak yes, when you're ready ;)

PR #983, opened 2018-02-21
Generic Foreign Key content type field autcompletion

Guilouf:
And also lighter syntax for creating FutureModelForm for GFK

**Usage**:

`models.py`

```python
class MyModel(models.Model):
    name models.CharField(max_length=300)
    content_type_field = models.ForeignKey(ContentType, on_delete=models.CASCADE, null=True, blank=True,
                                    limit_choices_to=models.Q(model='Model1') | models.Q(model='Model2')  
                                    ) 
    object_id_field = models.CharField(max_length=300, null=True, blank=True)
    content_object = GenericForeignKey('content_type_field', 'object_id_field')

    def __str__(self):
        return self.name
```

`forms.py`

```python
class GFKMyModelForm(metaclass=GenericForeignKeyModelFormMeta):
    content_type_field_name = 'content_type_field'  # attribute name from models.py
    object_id_field_name = 'object_id_field'
    view_name = 'autocomp'  # the view url name
    filter_queryset = [Model1, Model2]

    class Meta:
        model = Treatment
        fields = '__all__'
```

`view.py`

```python
class AutocompView(Select2QuerySetSequenceView):

    def get_queryset(self):
        # get the models direclty for the form object
        queryset_models = [model.objects.all() for model in GFKMyModelForm.content_type_models]  
        if self.q: 
            queryset_models[0].filter(alias__icontains=self.q)
            queryset_models[1].filter(name__icontains=self.q)

        # Aggregate querysets
        qs = QuerySetSequence(*queryset_models)

        qs = self.mixup_querysets(qs)

        return qs
```

Im aware that the small gain in syntax simplicity is largely counterbalanced by the increase of complexity on the source code, which is just a set of of ~ hacks because that's the only way i found without deeper changes in the source code.

I didn't code it the good way, but i think the modelForm interface i made for GFK is easier to use than the previous one, and the autocompletion of the content_type field is an useful feature.

For now it only support one GFK(content_type + object_id), but it only need a loop to support multiple GFK from one form.

### DIFF
--- a/src/dal/forms.py
+++ b/src/dal/forms.py
@@ -163,8 +163,8 @@ class FutureModelForm(forms.ModelForm):
         return self.instance
 
       
-  class GenericForeignKeyModelFormMeta(type): 
-    
+class GenericForeignKeyModelFormMeta(type): 
+
     def __new__(mcs, name, bases, dct):
 
         class GenericForeignKeyModelForm(autocomplete.FutureModelForm):
--- a/src/dal/forms.py
+++ b/src/dal/forms.py
@@ -166,14 +166,17 @@ class FutureModelForm(forms.ModelForm):
 class GenericForeignKeyModelFormMeta(type): 
 
     def __new__(mcs, name, bases, dct):
+        content_object = getattr(dct['Meta'].model, dct['gfk_name'])  # get the generic foreign key field from the model
+        ct_field = content_object.ct_field  # content_type field name
+        fk_field = content_object.fk_field  # object id field name
 
         class GenericForeignKeyModelForm(autocomplete.FutureModelForm):
-            content_type_models = dct['filter_queryset']  # models to be validated
+            content_type_models = dct['filter_queryset']  # models to be validated, needs to be there, readed from the view
             queryset_models = [model.objects.all() for model in content_type_models]
             _queryset_seq = autocomplete.QuerySetSequence(*queryset_models)
 
             # class attribute set from string
-            locals()[dct['object_id_field_name']] = GenericForeignKeyModelField(
+            locals()[fk_field] = GenericForeignKeyModelField(
                     queryset=_queryset_seq,
                     required=False,
                     widget=QuerySetSequenceSelect2(dct['view_name']),
@@ -182,14 +185,14 @@ class GenericForeignKeyModelFormMeta(type):
             class Meta(dct['Meta']):
                 widgets = {
                     # the content type field is set to hidden, and will be autocomplete in clean()
-                    dct['content_type_field_name']: forms.HiddenInput(attrs={'value': ''}),
+                    ct_field: forms.HiddenInput(attrs={'value': ''}),
                 }
 
             def clean(self):
-                content_type_obj = self.fields[dct['object_id_field_name']].content_type  # from GenericForeignKeyModelField
+                content_type_obj = self.fields[fk_field].content_type  # from GenericForeignKeyModelField
                 cleaned_data = super().clean()
                 # autocomplete the hidden content_type field
-                cleaned_data[dct['content_type_field_name']] = content_type_obj
+                cleaned_data[ct_field] = content_type_obj
                 return cleaned_data
 
         return GenericForeignKeyModelForm
--- a/src/dal/forms.py
+++ b/src/dal/forms.py
@@ -53,6 +53,7 @@ from django import forms
 
 from dal_queryset_sequence.fields import GenericForeignKeyModelField
 from dal_select2_queryset_sequence.widgets import QuerySetSequenceSelect2
+from queryset_sequence import QuerySetSequence
 
 
 class FutureModelForm(forms.ModelForm):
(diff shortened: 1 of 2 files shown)


jpic:
Hi @Guilouf,

Currently in Django, the recommended way is to instead make a model field (in this case: GenericForeignKey) subclass with a formfield() override that will return the default formfield for a model field.

Eventually, you could also add a as_view() method that could create a view type on the fly and return it, then you could do everything from the model field, even add as_url() so you can register the url name in it as well and have everything super dry.

What do you think of this proposal ?

Guilouf:
Hi @jpic, 
Sounds like a good idea, but i don't know exactly how it would be done and i can eventually see some drowbacks:
- Subclassing a `ModelField` means that the `models.py` will not be totally standard. However, i think DAL only concerns the "graphical interface" and should maybe not interfere with the models. Of course the `formfield()` method is dedicated to this "graphical" options and just return a form field instance, but maybe we should sacrifice a bit of DRY and then have something that would not enter in conflict with other representations like DRF for example.
- I don't know how to modify the other fields (content_type (`ForeignKey`) and object_id (`CharField`)) just with subclassing the `GenericForeignKey` field, without metaclass. However, i can improve my code with getting the `content_type` and `object_id` fields name from the `GenericForeignKey` field.

When you speak about the `as_view()` and `as_url()` methods, i understand it's about generate the `Autocomp` view from my example, but i don't see exactly how to do it just from the model field.
I just hope i haven't understood everything in the wrong way ;)

jpic:
DRF uses formfield too when displaying a form, and DAL is compatible with DRF formfield. Having your relation field's formfield return an autocomplete field will actually make you have autocomplete out of the box in DRF's filter form by default, without you having to specify it: https://github.com/yourlabs/django-autocomplete-light/issues/984#issuecomment-369075486

I actually have tried many ways to improve the situation by contributing to django, with signals, or new APIs, or new settings, always refused. I don't care anymore, adapters are a better solution.

How can a field generate a default view well it's actually simple:

```
class YourModel(models.Model):
   other = AutocompleteForeignKey('othermodel', search_fields=['whatever'], help_text='other model to link')

urlpatterns.append(YourModel.other.as_url())
```

Why ? Because contrary to what we believe before we actually read django models field code, a Model Field is not a DB field, a Model Field **has** a DB Field **and** a Form Field generation method. It seems to me that people sort of have lost of their sight how models were actually designed and you can figure it out by reading their code.

Anyway, I really don't care, because in adapters it's different: when you add the autocomplete adapter to a payload that was structured by a model adapter then the autocomplete adapter will be able to map field autocomplete adapters to the keys corresponding to model relations after finding the autocomplete view that goes with it. A completely different pattern, imho, a pattern for the future, as imagined by @mjtamlyn, Django core developer.

jpic:
Look i could troll django-dev mailing list by naively asking why does the ModelField has a formfield method, and why does Model does not have a form method and has to go through a factory. But that wouldn't even wake up anybody, I've tried to say it nicely, I don't think it would be welcome that I reiterate my thoughts in a trolly way.

jpic:
https://blog.yourlabs.org/post/171689946553/model-in-django-not-what-you-think-they-are

Guilouf:
I understood that a ModelField has a formfield() method which generate the formField of the actual form ( i thought before that the modelfield => formfield was done by form metaclass ).

I was worried about subclassing modelField by this kind of design: https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L839, where the modelField type can count.

For the view i agree there is something much more DRY to do.

What i still don't understand is how to go from the GFK that require **3** different model fields to just 1modelField. Is it possible for a model field to create more than one db record ?
If it's not possible then i don't see how to modify other fields (content_type field) from within a modelField.

Did you want to go from this:
```python
    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE, 
                                     limit_choices_to=models.Q(model='model1') | 
                                     models.Q(model='model2')) 
    object_id = models.CharField(max_length=300)
    content_object = GenericForeignKey('content_type', 'object_id')
```
to this ?:
```python
    dal_gkf = dal.models.GFKfield(...)
```
To be clear, on this [tutorial](http://django-autocomplete-light.readthedocs.io/en/master/gfk.html), the resulting form was nice but `content_type` field was not automatically filled (did i miss something?), and that's was one of the reason for this PR.

I didn't quite get what you where saying about adapters, but my design is maybe not so far from this, because it also map fields to others, ?

jpic:
@Guilouf would you please share some of your expertise on this PR ?
https://github.com/mjtamlyn/django-adapters/pull/43

jpic:
Also another approach here, where autocomplete views generate their widgets and form fields, another desperate attempt:

https://github.com/yourlabs/django-autocomplete-light/pull/589

I feel you ..

jpic:
Meanwhile I'm ok to merge this, as long as you support it, and if you add documentation, you can add tests too, this will help against another contributor breaking it.

Guilouf:
Yep, i will support it and add the doc/tests. We will have to think about the "interface", i mean this is ok for now but it does not handle multiple GFK, and it's not very nice:

```python
class GFKMyModelForm(metaclass=GenericForeignKeyModelFormMeta):
    gfk_name = 'gfk_model_field_name'  # attribute name from models.py
    filter_queryset = [Model1, Model2]
    view_name = 'autocomp'

    class Meta:
        model = Treatment
        fields = '__all__'
```
but something like this may be the goal for the "interface":
```python
class GFKMyModelForm(metaclass=GenericForeignKeyModelFormMeta):
    gfk1_model_field_name = DummyGFKObj(filter_queryset=[Model1, Model2], view_name='autocomp1')

    gfk2_model_field_name = DummyGFKObj(filter_queryset=[Model1, Model2], view_name= 'autocomp2')


    class Meta:
        model = Treatment
        fields = '__all__'
```
It handles multiple Gfk, the field names are written the fakely "subclassed" gfk, no need to pass them as strings. The `DummyGFKObj` is just there to wrap the data into an object that can be detected by the metaclass, could be just a tuple but an object is maybe nicer, and can be evolved along updates

I will take a look to the other PR you referenced, hopefully tomorow if got the time.

Guilouf:
Well, i finally find out that this PR of mine is a complete sh*t.
I was misslead because i confound fields in the form, then thinked there was missing feature then started to reinvent the wheel, which worked but in a very ugly way.

Anyway, i made this #995 which was inspired by our discution and i think is actually good, the test admin page is working, can you check it out ?

PR #990, opened 2018-03-10
Build single distfile

jpic:
**That's really not working very well, quite some bugs here, any better suggestion is welcome**

This is a config made from the one in yourlabs/facond.

The point is to compile our js into a single file with jquery and select2 in it, to bypass django's form media management system.


jpic:
@luzfcb this loads the right language file, but loading the language file fails because of a JS error.

Any idea please ?

luzfcb:
@jpic What is the correct way to build this?
I installed nodejs via https://pypi.python.org/pypi/nodeenv and i run `npm install`, but a got a lot of error messages

jpic:
Normally you only need to run sudo npm install -g yarn; yarn install inside the repo

jpic:
You don't need nodeenv, yarn install will populate the local directory's node_module directory.

If you want a crash course for npm: https://blog.yourlabs.org/post/171557820263/js-tdd-in-your-django-apps

gagarski:
Wouldn't there be a problem if the user includes another jQuery in addition to our bundle?

Maybe we should treat jQuery as external dependency: https://webpack.js.org/configuration/externals/ ?

gagarski:
BTW, what's the exact problem with Media in Django 2.0?

jpic:
There's a report that having a datetime widget and a dal widget on the same admin form would break js for one of the two.
Also an RFC in progress for Django: https://github.com/django/django/pull/9743
On django-webpack-loader I'm suggesting a completely different approach, what i think it should look like in the future, and supported by [adapters](https://github.com/mjtamlyn/django-adapters/pull/43): https://github.com/ezhome/django-webpack-loader/issues/141

jpic:
Fixed in 3.7.0.dev0

PR #995, opened 2018-03-15
Added support for automatically generated views for generic foreign key fields

Guilouf:
**Usage**: 
`forms.py`
```python
class TForm(autocomplete.FutureModelForm):
    test = autocomplete.GenericForeignKeyModelField(
        model_choice=[(Group, 'name'), (TModel, 'name')],  # Model with values to filter
        required=False,
    )

    class Meta:
        model = TModel
        fields = ('name',)
```
`url.py`
```python
urlpatterns.append(*TForm.as_url())
```

- The view to filter the result displayed by the select2 widget is now automatically created
- Support different fields to filter
- The generic foreign key field instanciation in the form needs less argument, it's more DRY

Tests in the admin page seems ok, it's runable as it. I still need to add more tests and doc, and check if there is no problem with model with multiple GFK, but should be ok.

Maybe classes will have to be moved to other package, i don't know what was the best place to put them.

### DIFF
--- a/src/dal_queryset_sequence/fields.py
+++ b/src/dal_queryset_sequence/fields.py
@@ -7,8 +7,11 @@ from dal_contenttypes.fields import (
 
 from django import forms
 from django.contrib.contenttypes.models import ContentType
+from django.conf.urls import url
 
 from queryset_sequence import QuerySetSequence
+from dal_select2_queryset_sequence.widgets import QuerySetSequenceSelect2
+from dal_select2_queryset_sequence.views import Select2QuerySetSequenceAutoView
 
 
 class QuerySetSequenceFieldMixin(object):
@@ -133,3 +136,29 @@ class QuerySetSequenceModelMultipleField(ContentTypeModelMultipleFieldMixin,
                 self.raise_invalid_choice(params={'value': val})
 
         return queryset
+
+
+class GenericForeignKeyModelField(QuerySetSequenceModelField):
+    """
+    Field that generate automatically the view for the QuerySetSequenceSelect2 widget
+    """
+    def __init__(self, *args, model_choice=None, **kwargs):
+        if model_choice:
+            self.model_choice = model_choice
+            models_queryset = [model[0].objects.all() for model in model_choice]
+            kwargs['queryset'] = QuerySetSequence(*models_queryset)
+
+        super().__init__(*args, **kwargs)
+
+    def as_url(self, form):
+        # if widget was given as kwarg, should have this url name !
+        url_name = '{}_autocomp_{}'.format(form.__name__, self.__class__.__name__)
+        if self.widget:  # add the widget if no widget kwarg
+            self.widget = QuerySetSequenceSelect2(url_name)
+
+        Select2QuerySetSequenceAutoView.model_choice = self.model_choice  # send to the view the model and filter list
+        AutoView = type('Autoview', (Select2QuerySetSequenceAutoView,), {})
+        # generate the class to work with multiple gfk (can't work on instance level)
+        AutoView.model_choice = self.model_choice
+        return url(r'^{}_{}_autocomp$'.format(form.__name__, self.__class__.__name__),
+                   AutoView.as_view(), name=url_name)
--- a/src/dal/autocomplete.py
+++ b/src/dal/autocomplete.py
@@ -57,6 +57,7 @@ if _installed('dal_queryset_sequence'):
     from dal_queryset_sequence.fields import (
         QuerySetSequenceModelField,
         QuerySetSequenceModelMultipleField,
+        GenericForeignKeyModelField,
     )
     from dal_queryset_sequence.views import (
         BaseQuerySetSequenceView,
--- a/src/dal/forms.py
+++ b/src/dal/forms.py
@@ -158,3 +158,13 @@ class FutureModelForm(forms.ModelForm):
             # saving of m2m data.
             self.save_m2m = self._save_m2m
         return self.instance
+
+    @classmethod
+    def as_url(cls):
+        """
+        Create a list of url patterns, to be called in url.py:
+        urlpattern.append(*ModelForm.as_url())
+        Iterate over the fields to call the as_url() method from the GenericForeignKeyField
+        """
+        return [value.as_url(cls) for key, value in cls.__dict__['declared_fields'].items()
+                if hasattr(value.__class__, 'as_url')]  # checks if its the right object
--- a/src/dal_select2_queryset_sequence/views.py
+++ b/src/dal_select2_queryset_sequence/views.py
@@ -4,6 +4,7 @@ from collections import OrderedDict
 from dal_queryset_sequence.views import BaseQuerySetSequenceView
 
 from dal_select2.views import Select2ViewMixin
+from queryset_sequence import QuerySetSequence
 
 from django.template.defaultfilters import capfirst
 from django.utils import six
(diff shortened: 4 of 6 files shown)


jpic:
Yes agreed perhaps a Select2GenericForeignKeyField could extend GenericForeignKeyField and override view class to use the select2 view ? Do you think that could work ?

Absolutely superb well done @Guilouf <3

Guilouf:
Thanks !
Hum yea that's might be possible but why use the `select2` view instead of the `select2_queryset_sequence` view ?
How the filtering on multiple models would work without `QuerysetSequence` ?

Guilouf:
Hum im afraid i don't understand:
- `GenericForeignKeyModelField` inherit from `QuerySetSequenceModelField`, why have a `Select2GenericForeignKeyField` that inherit from `GenericForeignKeyModelField` ?
- When you speak about the overriding view class, the view class generated by `GenericForeignKeyModelField` already inherit from `Select2QuerySetSequenceView`

I tested in the admin interface, it works also with model with multiple GFK. However i didn't success to run the unit tests, i have some error that i didn't much investigated yet:

```python
File "/usr/local/lib/python3.6/dist-packages/django/db/models/fields/related.py", line 606, in resolve_related_fields
    raise ValueError('Related model %r cannot be resolved' % self.remote_field.model)
ValueError: Related model 'select2_foreign_key.TModel' cannot be resolved
```
I will add the docs soon.

src/dal_queryset_sequence/fields.py
```diff
@@ -7,8 +7,11 @@
 
 from django import forms
 from django.contrib.contenttypes.models import ContentType
+from django.conf.urls import url
 
 from queryset_sequence import QuerySetSequence
+from dal_select2_queryset_sequence.widgets import QuerySetSequenceSelect2
+from dal_select2_queryset_sequence.views import Select2QuerySetSequenceAutoView
```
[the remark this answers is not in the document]
jpic: Currently, dal_queryset_sequence module is meant to couple dal and dqss, but there are branches of this repositories which use different js scripts and we'd like to keep it hacker friendly.

Please friend move dal_select2* from dal_queryset_sequence, one is presentational, the other is business logic.

I would like to merge it then and party for your welcome to the yourlabs hacker club contributors for the lulz and growing code and ponies

jpic:
Amazing design pattern !!!

Only one class to move and we're good to go if you want to maintain your patch in this repository .

src/dal_queryset_sequence/fields.py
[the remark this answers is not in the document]
Guilouf: Ok, i think i understood, im creating a `Select2GenericForeignKeyModelField` (with default select2 widget) and `GenericForeignKeyModelField` with no default widget.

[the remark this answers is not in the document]
Guilouf: I struggle for the "base"class, `GenericForeignKeyModelField`, which kind of widget should it handle ? If its meant to be used with a widget that accept url (`widget(url)`) like `QuerySetSequenceSelect2(url_name)` its ok. Same for the view, if its meant to accept something like `View.as_view(queryset=...)` like `Select2QuerySetSequenceView` then it works. But if the view or the widget don't fits, thene `GenericForeignKeyModelField` is useless. Dunno if im very clear tho..

src/dal_select2_queryset_sequence/fields.py
```diff
@@ -0,0 +1,34 @@
+from django.conf.urls import url
+
+from dal_queryset_sequence.fields import QuerySetSequenceModelField
+
+from queryset_sequence import QuerySetSequence
+
+from dal_select2_queryset_sequence.widgets import QuerySetSequenceSelect2
+from dal_select2_queryset_sequence.views import Select2QuerySetSequenceAutoView
+
+
```
Guilouf: Move and renamed QuerySetSequenceModelField

src/dal_queryset_sequence/fields.py
```diff
@@ -142,24 +140,27 @@ class GenericForeignKeyModelField(QuerySetSequenceModelField):
...
 
+        if isinstance(widget, type) and isinstance(view, type):  # check if they are classes
+            self.widget_obj = widget
+            self.view_obj = view
+        else:
+            raise AttributeError("Class object are required (not instantiated)")
+
         super().__init__(*args, **kwargs)
 
     def as_url(self, form):
```
Guilouf: Tried to handle custom widgets and views, but only if the wdget takes an url as argument and if the view can take a queryset argument. Maybe this is not a good idea, and i could just delete this class and just have  `Select2GenericForeignKeyModelField`

Guilouf: Ok cool then if you thinks its ok, `Select2GenericForeignKeyModelField` could inherit from `GenericForeignKeyModelField` and i have to update the docs in consequence. 
So you are ok with the fact that `GenericForeignKeyModelField` may not handle all kind of widgets ?

jpic: Of course, a form field is not always compatible with the widget class its passed, what would happen if you use a multiple select widget on a single choice field ? Nothing good. So for me it's ok.

Guilouf: Ok, then i polish it and it will be ready to be merged ;)

Guilouf:
Ok i added the docs i think its ready. Finally `Select2GenericForeignKeyModelField` did not inherit from `GenericForeignKeyModelField` because they have not so much in common.

jpic:
Perfect, can you rebase on master please ?

Guilouf:
Done, i think ;)

jpic:
Only one test broken for me:

```
$ tox -e py36-django20-sqlite-base
select2_generic_foreign_key/test_forms.py::GenericFormTest::test_initial FAILED                                                                         [ 27%]

E       django.urls.exceptions.NoReverseMatch: Reverse for 'select2_gfk' not found. 'select2_gfk' is not a valid view function or pattern name.
```

It seems like we're always generating a different url name, according to ./manage.py show_urls in test_project:

```
/select2_generic_foreign_key/TForm_140508816163840_autocomp	dal_select2_queryset_sequence.fields.AutoviewTForm140508816163840	TForm_autocomp_140508816163840
/select2_generic_foreign_key/TForm_140508816165464_autocomp	dal_queryset_sequence.fields.AutoviewTForm140508816165464	TForm_autocomp_140508816165464
```
Running again makes a different url:
```
/select2_generic_foreign_key/TForm_140111776965520_autocomp	dal_select2_queryset_sequence.fields.AutoviewTForm140111776965520	TForm_autocomp_140111776965520
/select2_generic_foreign_key/TForm_140111776967144_autocomp	dal_queryset_sequence.fields.AutoviewTForm140111776967144	TForm_autocomp_140111776967144
```
Do you think it's possible to make it somehow predictible ? ie. autocomplete_FormClassName_FieldName ?

Guilouf:
`autocomplete_FormClassName_FieldName` is what i wanted to achieve, but i didn't succeed to get the field name so instead i took the instance id which is unique and random.

Basically, to get the field name, we need to pass it from the `Form` to `Select2GenericForeignKeyModelField`:

```python
class TForm(autocomplete.FutureModelForm):
    test = autocomplete.Select2GenericForeignKeyModelField(
        model_choice=[(Group, 'name'), (TModel, 'name')],  # Model with values to filter
        required=False,
    )

```
I don't know how to do it, except explicitly pass the field name as an argument to `Select2GenericForeignKeyModelField`, which is ugly.

I have already research some means to get the name of the variable of where the instance goes, does not seems it's possible, i even exposed my method here: https://stackoverflow.com/a/49321993/5823489

The only way i see it might be possible is maybe create a metaclass for the form, before the field is instanciated, but again its quite ugly

Eventually, we can pass the `id` as an optional argument, so then by default it's random but if we want to specify it, like for passing the tests, then it takes the value we want. I think it might be the best solution, don't you ?

jpic:
Right, ok nevermind then don't waste your time investing into temporary hacks let's just wait for django-adapters which promises great introspection to become a thing.

Do you think a quick hack is possible in the test code to make it pass ?

Guilouf:
I think (didn't know how to launch the test -_-) i fixed it by adding the optional field_id argument.

jpic:
With command `tox -e py36-django20-sqlite-base` in the repo ;)

Guilouf:
I made a mistake in the url name, now the reverse url is working in test, but `self.assertHTMLEqual(result, expected)` is not working, there is missing `<options>` in the expected html, dunno why and struggle a bit understanding those tests.

Guilouf:
Strange, it was ok on my side and i got your latest commit ( "Use admin statics"). Now everything is completly f*cked up on my side, i will probably look at this tomorow.

jpic:
Try with the tig command:
2018-03-29-215156_1057x301_scrot
Apparently at some point you made a merge instead of a rebase.
I can squash and merge though ;)

PR #1000, opened 2018-03-26
Micro improvements for the create option

tonimichel:
1. Improved the german translation for the create option label text.
2. Added a special 'dal-create' class in case the select2 option represents the create option to instantly create a new item. (--> I think it is generally useful to have this class in order to distinguish a data item from the create option... now a concrete application can add its custom css to style the create option.)


jpic:
Why not add span.dal-create on the server side ? Given the state of the codebase, i would recommend avoid adding untested client side code.

Thanks for your contribution <3

jpic:
Sorry this didn't make it in 3.3.1, looking forward for your feedback .

PR #1002, opened 2018-03-28
Use admin statics

### LINKED ISSUE #959: Django 2.0 widgets may load jQuery after dal
In Django 2.0+, Django loads `jquery.js` instead of `jquery.min.js` when `settings.DEBUG=True`. See django/django@c19b56f which implements [Django #28377](https://code.djangoproject.com/ticket/28377) ([docs](https://docs.djangoproject.com/en/2.0/topics/forms/media/#form-media-asset-order)).

In debug mode, when `AdminDateWidget`, the default admin widget for `DateField`, appears after a dal widget, `select2.js` can't find jQuery (`Media.merge` ordering breaks?) and we also get `MediaOrderConflictWarning`, as mentioned in #788.

As a temporary workaround, I've overridden `AdminDateWidget` to always load `jquery.min.js`, even in debug mode. I imagine the fix should go in `Select2WidgetMixin`, thoughts?

jpic:
Does this fix #959 ?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "src/dal_select2/static/autocomplete_light/vendor/select2"]
-	path = src/dal_select2/static/autocomplete_light/vendor/select2
-	url = https://github.com/select2/select2.git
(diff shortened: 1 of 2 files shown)


jpic:
Merged in  35b0046

PR #1004, opened 2018-04-07
Getting placeholder and minimumInputLength from dal select

marianobianchi:
I was not able to get placeholder working until this change. The same problem was with minimumInputLength. This seems to fix both problems


marianobianchi:
I think these errors aren't from this PR. I understand that errors are from some python code not following pep8. But my changes are only JS code. What do you think? Will it be merged soon?

PR #1015, opened 2018-06-04
Set select2 container CSS class to :all:

hbielenia:
This PR sets `containerCssClass: ':all:'` option when initializing select2. In the generated element, this makes select2 use the same CSS classes as were set on original select element.

This fixes some select2 display bugs that are encountered with Bootstrap 4. Bootstrap requires the form elements to have `form-control` class, which is lost on a generated element without this fix. See example [here](https://github.com/select2/select2-bootstrap-theme/pull/72#issuecomment-381151068).


fauno:
Hi, I've found that setting `containerCssClass: ':all:'` prevents single select2 from showing the dropdown, can someone confirm?  it's working fine otherwise

PR #1027, opened 2018-08-06
Add Django 2.1 to the build matrix

browniebroke:
Django 2.1 has been released last week, so adding it to the build matrix.

This should help reproduce bugs like #1026 which is a problem only on Django 2.1+


jpic:
Thanks Kevin, you're the first in AUTHOR that has not contributed python code FYI :joy:

PR #1030, opened 2018-08-13
Fix ImportError with SELECT2_TRANSLATIONS in Django 1.x

### LINKED ISSUE #1029: Django SELECT2_TRANSLATIONS does not exists in 1.11
Commit 35b0046686cd883cf83f66f3cf2b2d77af520047 uses `SELECT2_TRANSLATIONS` from `django.contrib.admin.widgets` but they do not exists in [Django 1.11](https://github.com/django/django/blob/1.11.15/django/contrib/admin/widgets.py)

hugorodgerbrown:
Django2 introduced the SELECT2_TRANSLATIONS dictionary, and this
was imported into the dal_select2.widgets module, causing an ImportError
when run against any version of Django < 2.

Fixes #1029

### DIFF
--- a/src/dal_select2/widgets.py
+++ b/src/dal_select2/widgets.py
@@ -9,7 +9,11 @@ from dal.widgets import (
 
 from django import forms
 from django.conf import settings
-from django.contrib.admin.widgets import SELECT2_TRANSLATIONS
+try:
+    # SELECT2_TRANSLATIONS is Django 2.x only
+    from django.contrib.admin.widgets import SELECT2_TRANSLATIONS
+except ImportError:
+    SELECT2_TRANSLATIONS = {}
 from django.utils import six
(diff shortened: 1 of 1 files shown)


jpic:
I'm not sure, is this patch working for users on 1.11 ?

@todorvelichkov whats the issue exactly with the empty dict ?

todorvelichkov:
`i18n_file` will be empty => there will be no internationalization, while the old version of the code is loading it from `autocomplete_light/vendor/select2/dist/js/i18n/{}.js`

jpic:
Can we add a command perhaps to download the static files into a static folder for django 1.11 and be compatible with 2.0 maybe ? what's less harmful ?

jpic:
This just fixes an ImportError, please open new pr for new improvements.

PR #1042, opened 2018-09-07
Fixed a jQuery conflict bug in the widget code.

coredumperror:
The comments in the diff go into detail about what this is doing, but the gist of this PR is that DAL's method of including Django's copy of jQuery was interfering with other javascript that was expecting a newer version of jQuery to be installed. I removed the override of `admin/js/jquery.init.js`, since DAL's version was doing it wrong, and changed `autocomplete_list/js/jquery.init.js` to do it right. The widget code now adds the jQuery from Django and then immediately removes it from the global namespace, restoring whatever version of jQuery was installed above the DAL widget code.

I originally discovered this problem when using a DAL widget on a Wagtail Page form that also had a DateTimeField. `jQuery.dateTimePicker()` was becoming undefined as soon as I added a DAL widget, and that turned out to be because DAL was overriding the jQuery 3.2.1 provided by Wagtail with the jQuery 2.2.3 provided by Django.

### DIFF
--- a/src/dal/static/admin/js/jquery.init.js
+++ b/src/dal/static/admin/js/jquery.init.js
@@ -1,8 +1,0 @@
-var django = django || {};
-django.jQuery = jQuery
-
-var yl = yl || {};
-yl.jQuery = django.jQuery
-
-// this file also prevents django.jQuery.noConflict
-// there's got to be a better way ...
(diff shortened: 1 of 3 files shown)


coredumperror:
Awesome! Thanks for being so prompt.

PR #1056, opened 2018-09-28
Fix for querysetsequence v0.9

monim67:
- fixes for the breaking change of django-querysetsequence v0.9.
- adds xfail to test case test_initial from select2_generic_foreign_key.


monim67:
Any ideas about why it's failing now?

jpic:
Was solved upstream, thanks for your contribution

Issue #1070, opened 2018-11-12
Port fix for jQuery usage in af721ca to Django 1.11 compatible releases (> 3.2.10)


mandarg:
So af721ca54c75625d80b22007db70c93ca096b81c proved to be the definitive fix for "Select2: An instance of jQuery or a jQuery-compatible library was not found." type issues seen when using just `dal` and `dal_select2` with no other jQuery or Select2 modules. The issues seem to have been due to the fact that `$` was not being defined by either django-jquery or the `jquery.init.js` in this module, but the bundled `select2.full.js` expected `$` to be present. (I was halfway towards writing a fix before I found @coredumperror's much better one on master, so I'll let him weigh in in case I'm saying something wrong).

Anyway, the reason for this issue is – I'm still stuck on using Django 1.11, which is supported only by the 3.2.10 release, so I couldn't take advantage of the fix on master (which is only in 3.3 and upwards, which are [not supported for Django 1.11](https://github.com/yourlabs/django-autocomplete-light/pull/1064/files). 

So, to use it with Django 1.11, I changed some minor things in the patch and applied it to the 3.2.10 release [here](https://github.com/mandarg/django-autocomplete-light/commit/4bd04780061b36bc608b58dd07d40c0836a3eba6). Would you be open to cutting a new 3.2.11 release with that patch applied? (This will mean a new branch starting at the `3.2.10` tag). [Direct link to patch](https://github.com/mandarg/django-autocomplete-light/commit/4bd04780061b36bc608b58dd07d40c0836a3eba6.patch) that can be `git am`'d if it helps.

Let me know if you want any more work/input from me here, this is just to gauge the level of interest in a fix and 3.2.11 release. I'm fairly certain this also closes #824, and possibly #886 as well.

jpic:
I don't mind merging your patch: I don't target 1.11 compat myself, but I have nothing against it.

jpic:
You mean like maintaining a branch in Django 1.11 and another in Django 2.0 ? oh my :joy:

mandarg:
> You mean like maintaining a branch in Django 1.11 and another in Django 2.0 ? oh my 😂

Er sorry, I'm getting a little confused here. Were you going to apply my patch directly, or do you want me to do something else, like create a merge request against some branch?

I see that there is a `dj111` branch, but it seems to be fairly behind the 3.2.10 release.

```
$ git log --oneline `git merge-base dj111 3.2.10`..3.2.10 | wc -l
[...]

Issue #1058, opened 2018-10-05
Select2: incompatible with Django 1.11


Schnouki:
Version 3.3.2 of `dal_select2` depends on [`admin/css/vendor/select2/select2.css`](https://github.com/yourlabs/django-autocomplete-light/blob/3.3.2/src/dal_select2/widgets.py#L57). But this file doesn't exist in Django 1.11; in fact the whole `admin/css/vendor` folder is [missing](https://github.com/django/django/tree/1.11.16/django/contrib/admin/static/admin/css). It seems to be introduced in [Django 2.0](https://github.com/django/django/tree/2.0/django/contrib/admin/static/admin/css).

Would it be possible to fix that please?
Thanks!

jpic:
Any effort put into making DAL compatible with old versions of Django and Python should better be put into porting your project to Django 2 and Python 3 (paying your tech debt).

That said, contributions are welcome. Please just try to not duplicate the issue.

Schnouki:
FYI 1.11 is not just an old version of the Django, it's the current LTS.

jpic:
@Schnouki being LTS doesn't make it recent. In March 2018, Ubuntu 16.04 was old, even if it was LTS until 18.04.

Contributions for python 2 or django 1.11 are welcome, but don't count on me because from my opinion it's old versions.

raratiru:
>  pay your debt as early as possible instead of waiting the last minute, because as you wait, interest for your tech debt grow.

Cruel, indeed, but technically correct from all respects - especially for a free service of this level.

jpic:
Please re-open if occurs again on 3.4.0 which re-enables python 2 support

PR #1071, opened 2018-11-15
Update `dj111` branch to 3.2.10 release, add port of af721ca from master

mandarg:
Hi,

This is my first stab at a pull request for the issue mentioned in #1070. The approach here is:

- Rebase `dj111` onto tag `3.2.10` to get all the updates till 3.2.10 (since that's the last supported `dal` release for Django 1.11)
- Fix merge conflicts
- Put my fix at the tip of the (updated) `dj111` branch
- [Future, hopefully] Cut all 3.2.x releases from this branch, since someone on Django 2 will want >=3.3.x anyway

Does this approach sound correct in general? 

As for specific things – other than my fix, the only non-common commits are 8f9a5ab, 5f35557, and e45972d. I fixed merge conflicts roughly as things made sense to me, but it's probable that I made mistakes there, and would appreciate input from someone more familiar with the testing setup.


jpic:
Amazing effort, i've read it and have found no malware, so, should we release this in a new package like django-autocomplete-light-dj111 ? lawl :joy:

jpic:
Hey, did you try copying the staticfiles from Django 2.0 in like src/dal_dj111/static/admin ... ?

Then you could enable Django 1.11 support in INSTALLED_APPS or something ?

What do you think ?

Keep up the great energy you will save the days for many users :+1:

mandarg:
So I see three possible approaches here:

Given your comment:
> should we release this in a new package like django-autocomplete-light-dj111 ? 

Let's call this **Approach A** would be to release this in a completely different package, called `django-autocomplete-light-dj111`.

Some problems with this approach that I see:
- Documentation would have to be written about this separate package for Django 1
- Versioning would be somewhat odd. What would the first version of this package be?

> Hey, did you try copying the staticfiles from Django 2.0 in like src/dal_dj111/static/admin ... ? Then you could enable Django 1.11 support in INSTALLED_APPS or something ?

I have not tried this, but the end effect would be (if I understand your query correctly), something like:

```
# DAL for Django 1
INSTALLED_APPS = [ 'dal_dj1'...
```
and
```
# DAL for Django >=2
INSTALLED_APPS = ['dal'... 
```

right? Basically the package would be the same, but you'd have different apps in `INSTALLED_APPS` for Django 1. Let's call this **Approach B**.

Again, I see the following problems with this approach
- Extra documentation needs to be written
- The concept of separate installed apps for Django 2 v/s 1 when you're using the same underlying package is somewhat confusing

In my comment above, I have suggested **Approach C** – which is, you merge this MR into branch `dj111`and simply tag commit 2409a35aa6772b459cf966388556bd05625ed81e as `3.2.11`. This gets around the problems I see in the approaches above:
- A very minor change in documentation is required (specifying that Django 1 is supported only for 3.2.x
- Versioning makes sense, and we don't have to figure out any new versioning for the - `-dj1` package

So, my view is that we should use Approach C and just tag 3.2.11 and all future 3.2 releases off the `dj111` branch. Since you have [already stated](https://github.com/yourlabs/django-autocomplete-light/issues/1058#issuecomment-427659693) that 3.3+ releases won't explicitly support Django 1, this goes very well with it – if someone asks for Django 1 compatibility, you can simply point them to the `dj111` branch and say “Accepting pull requests” :)

Thoughts?

jpic:
LGTM

jpic:
Fixed by 3.4.1 release

Issue #620, opened 2016-03-07
'vendor' folder is missing


guettli:
In 3.0.1 I get 404-File-not-Found for the JS files located in the 'vendor' directory.

Yes, I have run "collectstatic" :-)

senoadiw:
+1, installed 3.1.0 with pip and getting 404 for the following:
* http://localhost:8000/static/autocomplete_light/vendor/select2/dist/css/select2.css
* http://localhost:8000/static/autocomplete_light/vendor/select2/dist/js/select2.full.js

The /static/autocomplete_light dir after collectstatic (no vendor dir created):
* autocomplete.init.js
* jquery.init.js
* select2.css
* select2.js

jpic:
Tried git submodule update --init ?

senoadiw:
@jpic comparing with 3.0.4 release on PyPI it looks like 3.1.0 is missing the vendor dir in src/dal_select2/static/autocomplete_light/

jpic:
Thanks !!​

Fixed in 3.1.1

I guess we need to find a curative solution against that, perhaps
automating releases ^^

Sorry for the delay.

Best

senoadiw:
@jpic thanks! I can confirm 3.1.1 release solves this issue.

guettli:
Yes, fixed in 3.1.1
Thank you very much!

andybak:
Seems to have reappeared in 3.3.2

Steps to recreate:

1. Download the 3.3.2 gz file from PyPi
2. Search for directories called "vendor"

There's only one and it's in dal_jal. However dal_select2/widgets.py line 57 references 'admin/css/vendor/select2/select2%s.css' % extra

andybak:
Yes. Just to need to think of a tidy way to handle forward and backwards compat.

How about keeping the "missing" files in a separate dal app: "dal_legacystatic" or similar?

The docs could then clearly state: "If you're using Django < 2.0 then add dal_legacystatic to installed apps.

andybak:
OK. Hopefully this does the trick: https://github.com/yourlabs/django-autocomplete-light/pull/1085

jpic:
Congrats @andybak for fixing the issue in master with #1085 :champagne: 

 Please @all someone could confirm the fix works for the with master ?

Or do you prefer a release right away ?

mpeuss:
@jpic 
We tested the fix and it seems to be working. When will you release a new version?

jpic:
[...]

PR #1085, opened 2019-01-23
Add dal_legacy_static app so vendored select2 css/js can be used with Django <2.0

andybak:
See #620


jpic:
Thank you very much Andy !

dandavison:
Thanks for the fix! Is there a release coming up soon that will include this?

jpic:
Well, I've asked in #620 if anybody wanted to try before release.

But meanwhile, users have opened issues about Django's select2 version to be a bit old, so that fact would favor completely reverting to the submodule system.

PR #1088, opened 2019-03-10
[WIP] Restore compatibility with querysetsequence==0.9

optiz0r:
This commit contains a partial fix restoring compatibility
with querysetsequence==0.9 by dropping use of a private api
in favour of a public one.

Instead of using the content-type id and object-id as the value
for select options, instead the queryset index and object id are
used. The `QuerySetSequence` maintains an ordered list of the
`QuerySet` objects it is composed of, and the index can be used
retrospectively to retrieve the model given the index and pk.

See also: https://github.com/percipient/django-querysetsequence/issues/21#issuecomment-432649275
Partial fix for #1050

### DIFF
--- a/src/dal_queryset_sequence/fields.py
+++ b/src/dal_queryset_sequence/fields.py
@@ -1,35 +1,39 @@
 """Autocomplete fields for QuerySetSequence choices."""
-
 from dal_contenttypes.fields import (
-    ContentTypeModelMultipleFieldMixin,
     GenericModelMixin,
 )
 
+
 from django import forms
 from django.conf.urls import url
-from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
 
 from queryset_sequence import QuerySetSequence
+
+import six
 
 
 class QuerySetSequenceFieldMixin(object):
     """Base methods for QuerySetSequence fields."""
 
-    def get_queryset_for_content_type(self, content_type_id):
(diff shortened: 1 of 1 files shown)


jpic:
Thank you for your contribution @optiz0r 

However, is it possible to fix the `tox -e checkqa` job ? Note that there were 4 failures that I just fixed in master so you might want to rebase.

optiz0r:
Found some issues with this in testing; not good to merge yet. (I was really abusing the PR to run the tests as they weren't running properly locally).

optiz0r:
Approach flawed, abandoning this PR.

jpic:
Why that ?

optiz0r:
commentary in #1050

jpic:
Harsh !
​
All my deepest respect for going through with your attempt @optiz0r 

I guess we're going to need a "get first queryset matching that content type" kind of method upstream in dqss after all

optiz0r:
Indeed. I updated the ticket on the QSS side too. Bit awkward having this conversation trailed across three separate threads, which is partly why I closed this one down.

jpic:
Sorry about that, but it's on my backlog to start being active on dal soon
and clean up some of the mess / make a release taking in account all recent
feedback, perhaps even migrating CI to gitlab so that we have use resources
from our CI server and have much more stable results from selenium ...I
mean: i know it's become messy and plan to do something about it ;)

PR #1093, opened 2019-04-30
Fix not iterable tag value

rastislav-chynoransky:
Hi,
there's a bug on rendering widget data of integrated `django-taggit` module.
`TagSelect2` class expects nested iterables but gets only flat list.

Thanks for your comments!


jpic:
Works for me !

Grats for your contribution !

Have a great day :+1:

PR #1096, opened 2019-05-13
Taggit Select 2 Bug

mmoallemi99:
if we want to edit an instance and fill tags with already added tags it causes exception:
Tag Object Not Iterable

Screenshot from 2019-05-13 16-37-03

so i overwrited inherited function format_value() , options() in TaggitSelect2 and now it's good to go

now it shows tags already added to object
Screenshot from 2019-05-13 16-37-33

Bug Fixed!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,6 +57,7 @@ documentation to the project.
 - Michał Sałaban
 - Mike Covington
 - Mislav Cimpersak
+- Mohammad Moallemi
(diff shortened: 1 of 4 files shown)


mmoallemi99:
why no merge?!

oxalorg:
This was already merged in another PR #1093 . Thanks for the code though helped me figure out my issue :)

Issue #1099, opened 2019-05-22
django-nested-admin NestedStackedInline/NestedTabularInline adding new row does not forward fields


akshenk8:
The issue is explained here with sample code to replicate it.
https://github.com/akshenk8/dal-issue-sample

jpic:
I think forward field support was not tested with django-nested-admin. Open for contribution !

jpic:
Nice, thanks for the feedback, merged

PR #1100, opened 2019-05-22
django-nested-admin NestedStackedInline/NestedTabularInline adding new row does not forward fields

akshenk8:
https://github.com/yourlabs/django-autocomplete-light/issues/1099

### DIFF
--- a/src/dal/static/autocomplete_light/forward.js
+++ b/src/dal/static/autocomplete_light/forward.js
@@ -131,7 +131,9 @@
             forwardedData,
             divSelector,
             form;
-        divSelector = "div.dal-forward-conf#dal-forward-conf-for_" +
+        divSelector = "div.dal-forward-conf#dal-forward-conf-for-" +
+                element.attr("id") + ", " +
+                "div.dal-forward-conf#dal-forward-conf-for_" +
                 element.attr("id");
         form = element.length > 0 ? $(element[0].form) : $();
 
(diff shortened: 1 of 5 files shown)


src/dal/static/autocomplete_light/forward.js
```diff
@@ -131,7 +131,7 @@
             forwardedData,
             divSelector,
             form;
-        divSelector = "div.dal-forward-conf#dal-forward-conf-for-" +
+        divSelector = "div.dal-forward-conf#dal-forward-conf-for_" +
                 element.attr("id");
```
jpic: Looks like a pretty breaking change, any chance to support both versions at once ?

akshenk8: done.

akshenk8: But can you tell me about what exactly would it be breaking ? 
As I have taken care of the id in JS, Widget and Tests.

jpic:
Well it will work on new projects but for projects that are going to
upgrade dal before nested admin for some reason it could be nice if dal
still works if possible otherwise it doesn't matter much

akshenk8:
@jpic does this work now or anything else has to taken care of ?

jpic:
Well if you're going to break compatibility with older nested-admin versions, it would be great to mention it somewhere. Also, set a minimal version with >= in setup.py for django-nested-admin

akshenk8:
I have set the minimal version number for django-nested-admin in test/requirements.txt and setup.py.
Sorry for the delays in the fix.

jpic:
Nice, thanks !

PR #1105, opened 2019-07-09
Python 2 support (mostly)

jpic:
This patch adds kind of support for python2:

- select2 is back as a submodule, django 2 is not required anymore, and we can better follow releases from select2
- language code detection does not benefit from lru_cache in python2, but PR welcome
- a few tests still don't pass on python2, because dicts are not ordered

Besides those minor drawbacks it looks like this would work on python / dj111.

Would anybody like to test prior to merge and release ?

Remember to use git clone --recursive


jpic:
This patch adds support for Python 2 and Django < 2.
Would anybody like to test prior to merge and release ?
Otherwise will release in a month or so.

PR #1106, opened 2019-07-18
Delete select2 submodule from git index

jorrit-wehelp:
It's still the index, but not .gitmodules (got removed a while back, should now be using the select2 from Django 2's contrib.admin).

(This causes errors when switching back 'n forth between older branches/versions which do still have the submodule)


jpic:
Apparently the submodule is coming back to re-enable python 2 support ... https://github.com/yourlabs/django-autocomplete-light/pull/1105

PR #1107, opened 2019-07-18
Travisci fixes

jorrit-wehelp:
- fixes xvfb errors, cause by change in travisci environment
- workaorund for flake-docstrings error in testqa
- workaorund/fix line length erros from testqa

Since the lines causing the flake8 errors had already been merged and released I did the easiest thing and made 88 the new limit (borrowing from black's defaults), but that is debatable :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,9 @@ install:
 - travis_retry pip install -U pip
 - travis_retry pip install tox
 - travis_retry pip freeze
+services:
+  - xvfb
 before_script:
-- "export DISPLAY=:99.0"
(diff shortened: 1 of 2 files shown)


jorrit-wehelp:
The actual tests still fail, took a quick stab at replacing phantomjs with firefox (since phantomjs support has been removed from splinter). This runs into compatibility problems with travis's firefox and selenium geckodriver versions, so would require some more trail-and-error work.

But that is probably also better done in a separate branch/pull request.

jpic:
Nice, thanks !
You might want to try using circleci, many of us have spent *a lot* of time trying to make this work with travis again

PR #1109, opened 2019-07-25
Replace phantomjs with firefox

jorrit-wehelp:
This gets the liveserver/splinter/selenium based tests working again on TravisCI.

Did add a python 3.6 test run, just to show it can run without failures.
The 3.5 run shows 1 error, looks like dictionary/hash ordering problem in the test itself, so unrelated to this pull requests change (better fixed separately).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,50 +1,39 @@
 sudo: false
+dist: xenial
 language: python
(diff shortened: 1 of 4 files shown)


jpic:
Amazing

PR #1118, opened 2019-09-26
Highlight select field with error to match Django style

tchatow:
Since the Django generated select box is hidden with select2, the red border style does not appear when there is an error. This fix adds that style the select2 box.


jpic:
Nice one, thanks !

jpic:
Released in 3.5.0

PR #1121, opened 2019-10-18
Fix Compatibility issue with Django 3.0 due to django.utils.six

### LINKED ISSUE #1115: Compatibility issue with Django 3.0 due to django.utils.six
Django 3.0 scheduled for December removed Python 2 compatibility APIs used by django-autocomplete-light (django.utils.six, django.utils.encoding.python_2_unicode_compatible).

See https://docs.djangoproject.com/en/dev/releases/3.0/#removed-private-python-2-compatibility-apis
It would be wise to switch now to the direct use of six.

Mogost:
Fix #1115

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ setup(
     long_description=read('README'),
     license='MIT',
     keywords='django autocomplete',
+    install_requires=['six'],
     extras_require={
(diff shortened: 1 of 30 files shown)


jpic:
Released in 3.5.0

PR #1124, opened 2019-11-14
Initialize all items on DOMNodeInserted and don't exclude prefix

dwasyl:
This minor JS change enables DAL to work with the `jquery.formset.js` plugin that is commonly used. This PR reverses a change made a few versions ago adding more specificity to the `initialize`.


jpic:
Released in 3.5.0

Issue #1135, opened 2020-01-14
Regression formset bug introduced from upgrading 3.2.10 to 3.5.10.


jackton1:
BEFORE:

![Fix](https://user-images.githubusercontent.com/17484350/72379649-d3f7df00-36e1-11ea-98cb-1073f92f22b7.gif)

AFTER:

![Jan-14-2020 15-11-40](https://user-images.githubusercontent.com/17484350/72378903-51baeb00-36e0-11ea-9310-09a79f8dc227.gif)

VenturaFranklin:
For me cloning formsets with DAL fields still worked on v3.3.5, upgrading to 3.4.1 killed this and it sill does not work in 3.5.0. So something between 3.3.5 and 3.4.1

Specifically, running the code to allow dynamic add/remove of formsets using jquery: https://medium.com/all-about-django/adding-forms-dynamically-to-a-django-formset-375f1090c2b0

I had to make a few changes to get that tutorial to work, my final cloneMore function is below and this works fine with v3.3.5:
```javascript
function cloneMore(selector, prefix) {
    var newElement = $(selector).clone(true);
    var total = $('#id_' + prefix + '-TOTAL_FORMS').val();
    newElement.find(':input:not([type=button]):not([type=submit]):not([type=reset])').each(function() {
        var old_name = $(this).attr('name');
        if (old_name) {  // This is needed because DAL doesn't have name?
            var name = $(this).attr('name').replace('-' + (total - 1) + '-', '-' + total + '-');
            var id = 'id_' + name;
            $(this).attr({'name': name, 'id': id}).val('').removeAttr('checked');
        }
    });
    newElement.find('label').each(function() {
        var forValue = $(this).attr('for');
        if (forValue) {
          forValue = forValue.replace('-' + (total-1) + '-', '-' + total + '-');
          $(this).attr({'for': forValue});
        }
    });
    newElement.find('span').each(function() {
        // This is needed because DAL will be duplicated otherwise
        $(this).remove();
    });
    total++;
    $('#id_' + prefix + '-TOTAL_FORMS').val(total);
    $(selector).after(newElement);
    updateConditionRow();
    return false;
}
```
The two sections I had to add are noted above. Removing the section below does not affect the error at all: 
```javascript
newElement.find('span').each(function() {
        // This is needed because DAL will be duplicated otherwise
        $(this).remove();
    });
```

[...]

PR #1136, opened 2020-01-31
Demo improper clone of last object instead of empty form

VenturaFranklin:
@jpic I was able to recreate the issue shown in #1135 and the issue that I was having in the test project. I would say it is not an actual bug, but an improper cloning. 

Specifically, if you clone the last element in the formset list and NOT the empty formset as demoed in [select2_outside_admin.html](https://github.com/yourlabs/django-autocomplete-light/blob/master/test_project/select2_outside_admin/templates/select2_outside_admin.html) you will run into this issue. I show both the wrong way and the right way in this PR. You don't really need to merge this, but maybe reject and point to the right way to do things...


jpic:
I see, pretty funny sequence of event:

2020-03-18-232443_636x160_scrot
2020-03-18-232453_626x157_scrot
2020-03-18-232459_625x236_scrot

Did you find your way out with the empty formset ? I think it's better to clone from there, that's also how the admin deals with dynamic formsets as such it's going to be the preferred / most supported way from both django, admin and dal.

VenturaFranklin:
Yes! Empty formset is the way to go, but there are tutorials floating around that use the copying of the last element. I suggest in documentation highlighting that this will not work and you MUST use the empty formset to get it to work.

jpic:
True that, we should have a micro external app reproducing dynamic formsets from outside the admin (but exactly the same way to be compatible) like we already have with django-addanother, @jonashaag please share your thought

PR #1147, opened 2020-03-18
Remove jQuery loading complexity

jpic:
Side effect: removes whatever django is trying to protect users from with setting noConflict (allowing users to load several versions of jquery at the same time ?)


jpic:
Also published in : django-autocomplete-light==3.6.0-dev

Please try it out, it is expected to break fix in theory, but to fix things in practice, regarding jquery et al.

micjabbour:
I tested the published package on PyPI and it seems to work on my machine.  However, I noticed some differences in the [code in the package](https://files.pythonhosted.org/packages/a1/5d/84e7bfd32a14108890178ecf98d2e94281de7f91b747e5e6c1b72cb04a4a/django-autocomplete-light-3.6.0.dev1.tar.gz), does it use a different revision from this PR? For example:

- `autocomplete_light/autocomplete.init.js`, `autocomplete_light/forward.js`, `autocomplete_light/select2.js` have been modified in the PyPI package to use `$ || jQuery || django.jQuery` instead of [`yl.jQuery`](https://github.com/yourlabs/django-autocomplete-light/blob/jquery/src/dal/static/autocomplete_light/autocomplete.init.js#L123).
- The reference to `'autocomplete_light/jquery.init.js'` in the `media` function for `Select2WidgetMixin` has been [removed in this PR](https://github.com/yourlabs/django-autocomplete-light/pull/1147/files#diff-1fff4760d1a64d470685017c722100f3L88), but it still exists in the PyPI package (I guess it should be removed there as well).
- The new file `dal/static/admin/js/jquery.init.js` (which overrides the built-in Django-admin file) has been re-added in the PyPI package, but it is not available in this PR.

Am I missing something here?
Thank you very much

jpic:
Fixed in 3.7.0.dev0

PR #1158, opened 2020-05-19
remove calculated widths

nad2000:
select2 erroneously calculates the width of the elements. It has been a knows issues for a while: https://github.com/select2/select2/issues/208 It's not resolved, but there is a workaround:

```js
$(element).select2({
  ...
  width: null
  ...
});
```

If there is a need to readjust the width It should be done vis CSS or dynamically after completion of rendering.

This is how the UI looks with ModelSelect2 before:

screen_before000

After:

screen_after

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -80,6 +80,7 @@
             templateResult: result_template,
             templateSelection: selected_template,
             ajax: ajax,
+            with: null,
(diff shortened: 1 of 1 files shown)


nad2000:
@jpic, my bad I misspelt "width".
Also, it appears that **width: '100%'** works the best in this case. 
Can you pls reopen the PR? Cheers!

barclaey (commented):
Please fix this to: 
width: $element.attr('width') || '100%',

PR #1157, opened 2020-05-16
Fix jquery loading

Erhan23:

Issue #1156, opened 2020-05-13
dal incompatible with jQuery-File-Upload


daltschu:
I have a form with both file fields, foreign keys (ModelChoiceField) and m2m.
In the template I want to use jQuery-File-Upload [https://github.com/blueimp/jQuery-File-Upload]
to upload the files and dal for the fk and m2m.

I have a javascript/jQuery error on loading the page:
TypeError: $("#myfileupload").fileupload() is not a function
and the file upload won't work.

When I modify the view so that it's not using dal.autcomplete, the file upload works.
Did anyone got this problem before?

Instead of using dal for this view, I moved all the data needed (less than 300 instances) from the model underlying my fk and m2m in a javascript array with the template and used jQuery select2 and it worked fine.

I have dal version 3.5.1 with django 2.2.5 and blueimp-file-upload version 10.24.0

jpic:
This kind of error means that the plugin is not loaded into jquery, perhaps it was loaded before than another jquery was loaded on `$` ?

Issue #1153, opened 2020-04-28
select2.js duplicate with select2.full.js


PaleNeutron:
I got double select2 span in my admin inline if I use dal in it's parent page.

They are `http://127.0.0.1:8000/static/autocomplete_light/select2.js` and  `http://127.0.0.1:8000/static/vendor/select2/dist/js/select2.full.js`, if I block one of them, everything works.

How to resolve the problem?

PaleNeutron:
I can fix it by update my Django version from 2.1.14 to 2.2.12.

Both `selec2.full.js`  are  still loaded in the page, but this time, they works well, no duplicated selects.

So, what's the problem? how can I resolve it if I downgrade back to Django version 2.1.14?

jpic:
Honestly I'm not sure anymore, I can't seem to find a recipe that works
with all use cases.

As such, anything works for me really, as long as it passes the selenium
tests which do test admin inline.

Issue #1143, opened 2020-03-05
3.5.1 new Media order is not safely compatible with ModelAdmin.autocomplete_fields


raratiru:
In the following setup, one inline uses `autocomplete_fields` the other inline uses DAL. 

It works in 3.5.0. In order to work in 3.5.1. the `Media` meta-calss should also be defined in the inline that uses `autocomplete_fields`. 

```
class WithDALForm(forms.ModelForm):
    field_with_DAL = forms.ModelMultipleChoiceField(
        ...
        widget=autocomplete.ModelSelect2Multiple()
    )

    class Media:
        js = ("//code.jquery.com/jquery-3.4.1.min.js",)


class WithDALInline(admin.StackedInline):
    form = WithDALForm


class WithAutocompleteInline(admin.StackedInline)
    autocomplete_fields = ('field_1', 'field_2')


class Admin(admin.ModelAdmin):
    inlines = ('WithAutocompleteInline', 'WithDALIline')
```
-----

What follows is the order of the scripts in each case:

**3.5.1 with admin autocomplete (not working):**
```
<script type="text/javascript" src="/static/admin/js/vendor/jquery/jquery.js"></script>
<script type="text/javascript" src="/static/autocomplete_light/jquery.init.js"></script>
<script type="text/javascript" src="//code.jquery.com/jquery-3.4.1.min.js"></script>
<script type="text/javascript" src="/static/admin/js/vendor/select2/select2.full.js"></script>
<script type="text/javascript" src="/static/vendor/select2/dist/js/select2.full.js"></script>
<script type="text/javascript" src="/static/admin/js/vendor/select2/i18n/en.js"></script>
<script type="text/javascript" src="/static/vendor/select2/dist/js/i18n/en.js"></script>
<script type="text/javascript" src="/static/admin/js/jquery.init.js"></script>
<script type="text/javascript" src="/static/autocomplete_light/autocomplete.init.js"></script>
<script type="text/javascript" src="/static/admin/js/core.js"></script>
<script type="text/javascript" src="/static/admin/js/inlines.js"></script>
<script type="text/javascript" src="/static/admin/js/autocomplete.js"></script>
<script type="text/javascript" src="/static/autocomplete_light/forward.js"></script>
<script type="text/javascript" src="/static/admin/js/admin/RelatedObjectLookups.js"></script>
<script type="text/javascript" src="/static/autocomplete_light/select2.js"></script>
[...]

PR #1162, opened 2020-06-09
Fixed JS load order issues

danielmorell:
This PR seeks to solve the same problem as PRs #990, #1147, and #1157.

I could not find any testing / contributing documentation, so please advise if there are any issues with this PR, I can adjust. I spun up the `test_project`. However, I am not sure what the results are supposed to be. Most things appeared to be working.

### Here is an overview of what I did.

**❗ Update: The solution described here has been rewritten. [See the update below](https://github.com/yourlabs/django-autocomplete-light/pull/1162#issuecomment-642289913).**

I initialized a NPM project including `select2` as a dependency.

From there, I used a number of JS build tools to bundle all the autocomplete files into a single JS file called `autocomplete.js` and a minified version `autocomplete.min.js`.

The complicated part was handling i18n. I created a build script `select2.build.js` in the project root. It does two things. 

First, it collects all the Select2 language files, then writes them to `src/dal/static/autocomplete_light/i18n/`. It does not just copy each language file, it wraps the entire contents of each file in a function `dalLoadLanguage()`. The function takes a single argument `jQuery`. This allows us to call the function during the initialization.

https://github.com/danielmorell/django-autocomplete-light/blob/e717499a4efe69446aaa5a1141cd4c49531c6ac0/src/dal/static/autocomplete_light/autocomplete.init.js#L71-L81

It also registers and fires a custom event, `dal-language-loaded`. This is important since the file may be loaded after the main `autocomplete.js` bundle. I added an event listener that will call the `dalLoadLanguage()` function when `dal-language-loaded` is fired.

The last thing `select2.build.js` does is concatenate all the JS files into `autocomplete.js`. It, does not attempt to perform any optimization or modulization on the code. It simply write them all in order into one file.

Finally, `autocomplete.js` is minified using UglifyJS.

### The following build steps are now required.

**Install JS dependencies.**
```
$ npm install
```

**Build JS files.**
```
$ npm run build
```


jpic:
Well that's a serious contribution given that it's tackling what looks to be the pain point of thousands of users who are going to really appreciate to see the length you went through.

One question thought: do you think autocomplete.init.js and jquery.post-setup.js are still necessary ? I think they would be good to delete, they were mostly fighting the fact the whole JS scripts where served separately weren't they ?

As of other tests you can make, there's the selenium tests you can run with tox (`tox -l` to list envs, then use like `tox -e base-py38-django30-sqlite`, that should mostly work, except something with related with django-querysetsequence on Django 3 but there's an open discussion about it in their issues and they might as well already have released a version that now supports Django 3. 

Also installation in a fresh project from scratch following the installation tutorial can help finding release issues if any.

Keep up the great work :tophat:

danielmorell:
Thank you for the extra info! It will prove very helpful.

I am currently digging through the test results from Travis CI.

I have a quick question. There are some app models with no records in the sqlite DB. Is that intentional? Should I populate these models when testing or is that handled when the tests are initialized? They are... 
   1. `forward_different_fields/tmodel`
   1. `linked_data/tmodel`
   1. `rename_forward/tmodel`
   1. `secure_data/tmodel`
   1. `select2_list/tmodel`
   1. `select2_nestedadmin/tmodelone`
   1. `select2_one_to_one/tmodel`
   1. `select2_taggit/tmodel`
   1. `taggit/tag`

jpic:
It depends, some example apps have tests, others have tests that are not run anymore since Django 2, and some of them might even not have tests at all. From tox.ini:

```
    django{20,21,master}: py.test -v --cov --liveserver 127.0.0.1:9999 {posargs} secure_data rename_forward select2_foreign_key select2_generic_foreign_key select2_list select2_many_to_many select2_one_to_one select2_outside_admin select2_taggit custom_select2 select2_nestedadmin
    django1{8,9,10,11}: py.test -v --cov --liveserver 127.0.0.1:9999 {posargs}    
```

Now, each app should have their own `test*.py` files, all of them had test at some point (around the initial release DAL3), some of them have unit tests, most of them inherit from the selenium test story which would require initial models I suppose, and they test selection from the main form (without prefix) and from an inline.

jpic:
Great job overall, I'm not really sure why registering functions is necessary, but updating t_select2.js should tell you how to update the javascript override documentation.

At first look registering functions didn't seem so nice to me, but if there's a performance advantage then it's a small cost. Users often report performance issues when they have a lot of autocompletes in a page, so the benefit outweights the cost, and it's not going to be the first BC break in 8 years :joy:

I use npm, webpack and babel a lot, I have nothing against it, it's just I couldn't get select2 to play nice with it.

Apparently I'm fixing the django-querysetsequence issue this weekend, so hopefully all tests should be passing next week.

danielmorell:
The reasoning behind registering the functions is it ensures 100% predictable execution no matter the load order of the JS files, or the existence of jQuery in the lexical environment. In fact you could dynamically pull in a new JS file with a custom function long after the page has been rendered, and as long as you call `yl.registerFunction()` it will work. Personally, it smells right. 

I have been thinking about how to make it work without causing a breaking change. The most obvious way I can think of is to use a new event for function registration, and use the original `autocompleteLightInitialize` event to support previous custom functions. We can keep support for the old method until it makes sense to introduce a breaking change.

jpic:
Breaking BC is fine, particularly to fix such a widely spread problem.

If you want BC, then perhaps make a getFunction that will first check in
the registry, otherwise try to import from document.

To change documentation, it's just text files in the docs directory ;)

danielmorell:
I removed the breaking changes, updated the docs, commented my code, and updated `t_select2.js`.

I have done everything I think I need to. Please take a look over everything and let me know if the docs are clear, and if I need to update anything else.

jpic:
Excellent work, the documentation is better than ever too !

I'm going to run the tests locally on your branch after I make the django-querysetsequence patch this weekend, if all tests pass then we're good for a release !

danielmorell:
Here is a list of issues that I believe are related to this PR. There are some that identify the core bug, and some that I think may be related, and we may want to check back on the latter to see if the issue still exists.

**Issues that should be resolved**

- #1143
- #1137
- #1079
- #1082
- #1032

**Issues that may be related**

- #1156
- #1153
- #1110

danielmorell:
@jpic Have you been able to run this PR branch through tests? If any tests failed let me know the details, and I will be happy to work on resolving the issues.

danielmorell:
@jpic I just want to circle back on this PR and see if you had a chance to run tests?

jpic:
Hi, sorry for the delay, well npm install and npm run build went well, it seems to work, but selenium tests are not passing, not really starting anything besides opening the URL. Have you had more luck ?

jpic:
I'm still working on this, but it seems your system works, except in the example "outside the admin", where it seems to only work for the inline autocomplete, but not for the normal autocomplete which is pretty weird . Do you think you can fix that ?

I got the PR open on django-querysetsequence but apparently there's still a small bug I need to squash myself prior to releasing this

jpic:
I got our problems fixed, but the case here is that Django < 3.0 doesn't seem to work with your PR, would you like to take a look at that ?

Again , sorry for the delay

jpic:
I could get the test suite to work:

- removed jQuery from scripts executed in our selenium story code
- removed a test about select2 loading which is now irrelevant

I merged this in the command line.

Grats !

jpic:
Ok we have an actual problem, it's with the "Linked data" example, it has the following example script:

```
$(document).ready(function() {    
    $(':input[name$=owner]').on('change', function() {    
        var prefix = $(this).getFormPrefix();    
        $(':input[name=' + prefix + 'test]').val(null).trigger('change');    
    });    
});    
```

But that's not possible anymore: we don't have a $, we don't have $.getFormPrefix anymore.

Do you think you can find something for that ?

danielmorell:
I will create a new PR that addresses that. I will check to see if there are any others as well.

helderco:
This broke my customization. I had something like this:

```
(function ($) {
  $(document).on('autocompleteLightInitialize', '[data-dependant-fields]', function () {
      var $element = $(this);
      // my custom behavior
    },
  );
})(yl.jQuery);
```

The issue here is that I don't want to override the select2 init completely, I just want to add to it. 

I can do something like this:

```
document.addEventListener('dal-init-function', function () {
  var $ = django.jQuery;
  $('[data-dependant-fields]', this).each(function () {
      var $element = $(this);
      // my custom behavior   
  }
});
```
Which works initially, but **not when nodes are inserted after page load**.

The MutationObserver calls initialize(), which is where the previous event was being called, so it worked. But with that event being deprecated, how do I fix this?

Is it better to keep a custom event in initialize(), or allow to **register multiple callbacks**? That would be cool as I could also get the $ via callback parameter. Or is there something else I should be doing now?

helderco:
> Ok we have an actual problem, it's with the "Linked data" example, it has the following example script:
> 
> ```
> $(document).ready(function() {    
>     $(':input[name$=owner]').on('change', function() {    
>         var prefix = $(this).getFormPrefix();    
>         $(':input[name=' + prefix + 'test]').val(null).trigger('change');    
>     });    
> });    
> ```
> 
> But that's not possible anymore: we don't have a $, we don't have $.getFormPrefix anymore.
> 
> Do you think you can find something for that ?

Indeed, essentially this is what I need solving. If you could register multiple init callbacks, we could put this code there. It would solve, not only having $, but new elements detected by the MutationObserver.

Issue #1171, opened 2020-07-24
Using data-html with ability to create a new option displays [object Object]


mitchelljkotler:
If you set your select2 widget to use html by setting the "data-html" attribute to true, and also have the ability to create new options, the new option will show as `[object Object]` instead of rendering the text properly.  The offending code is here:

https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal_select2/static/autocomplete_light/select2.js#L11-L32

I am not sure why you need to wrap the text in a span if the item is being created, but if you are using HTML, the `template` function will return a span object, which you are then trying to insert as text into the second span.  Changing the text to HTML works, and does not break the non-HTML case, although I am not sure if that's an ideal solution or if it should be detecting this case and only inserting one wrapping span.

madison-ookla:
I am also experiencing this issue! I guess a workaround for me would be a way to apply a class to the query results container once it gets created - is there a way to do this that anyone knows of? It doesn't look like [overriding the JS as they describe](https://django-autocomplete-light.readthedocs.io/en/3.5.1/tutorial.html#overriding-javascript-code) works the way I was expecting it to

syserr0r:
I'm also experiencing this.

I've submitted a PR that should fix this based on your comments @mitchelljkotler

jpic:
Hi all,

Please try PR #1181 by @syserr0r and let us know if that works for you prior to releasing.

Thank you deeply :tophat:

jpic:
@madison-ookla sorry about that, it seems the docs are not updated, see the change in docs in a2705b40c4f92a3ed0da949fe4a9786ee6e86660

(RTFD builds fine, digging to figure the docs problem)

jpic:
@madison-ookla my bad, the link you are showing from the docs is from version 3.5.1, but 3.6.0 was released on July 6, the right link is fine:

https://django-autocomplete-light.readthedocs.io/en/master/tutorial.html#overriding-javascript-code

madison-ookla:
@syserr0r @jpic I can confirm that when I edit the files in my container to reflect the changes of #1181, the create object tab behaves as expected! Although this creates a new problem for me, as I need to find some other way to modify the HTML of the `Please enter <x> or more characters` section of the dropdown...

jpic:
@madison-ookla thank you ! It was merged and will release by the end of the week.

If you want to wait a couple of months, I'm pushing this to production soon:

[...]

PR #1181, opened 2020-09-14
Potential fix for #1171

syserr0r:
Fix for displaying [object Object] when data-html is true and object creation allowed, see ticket #1171

Untested, going purely from ticket comments.

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -21,11 +21,11 @@ document.addEventListener('dal-init-function', function () {
 
         function result_template(item) {
             var text = template(item.text,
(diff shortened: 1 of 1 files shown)


jpic:
Ok but in the comments you say that you are experiencing the same error, does that mean this patch works for you and may I pull your leg shamelessly and ask how to reproduce so at least I can verify myself prior to releasing ?

syserr0r:
Yes, as described in #1171 set data-html="true" while allowing object creation. If you need more information then please do let me know.

jpic:
Thank you, can you briefly try locally with your patch ?

jpic:
Confirmed working by another user !

Thank you deeply for your contribution !

Keep up the great work :tophat:

PR #1182, opened 2020-09-22
Fix autocomplete dropdown not work in production environment

MarkRunWu:
The autocomplete dropdown menu become single choice menu due to the following two issues:
1. `autocomplete_light.min.js`  is out date not the latest one, so i run `yarn build` to resync the two versions.
2. `vendor/select2/dist/css/select2%s.css` path is only valid in local server or host in django but the `vendor/select2/dist/` doesn't exist once export by `python manage.py collectstatic`, correct to use the exported static path.


jpic:
Nice one, thank you :tophat:

PR #1183, opened 2020-10-09
use backports.functools_lru_cache for lru_cache if available

pawl:
I'm working on a project that is still on python <3.3 and I'm seeing a lot of the "Python2: no cache on get_i18n_name until contribution" warnings. It seems like the best way to get rid of these warnings is to use `filterwarnings`. However, it would be nice to still be able to use the `get_i18n_name` caching.

This PR allows python <3.3 users to get rid of the warning and enable the `get_i18n_name` caching by using `lru_cache` from `backports.functools_lru_cache` if it's available.

### DIFF
--- a/src/dal_select2/widgets.py
+++ b/src/dal_select2/widgets.py
@@ -2,8 +2,12 @@
 
 try:
     from functools import lru_cache
-except:
-    lru_cache = None
+except ImportError:
+    # py2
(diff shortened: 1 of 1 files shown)


jpic:
Thank you for your contribution.

Releasing as part of 3.7.0

Issue #1151, opened 2020-04-03
[Sugestion] A View that returns a queryset with opt-group by that models field


epoiate:
Like a Select2GroupListView and Select2QuerySetView mixed togher.
I've done it at work and know it's possible but this would be my first OSS contribution so I would need some help setting up the tests and etc.

jpic:
Sure, do you have working code already ? Can we see it ?

epoiate:
Here is the pull request #1186

PR #1186, opened 2020-10-14
Implementing a grouped queryset

epoiate:
Enables Select2 to group options by a related model #1151


jpic:
That's fine for me, do you want to add a test to ensure that nobody breaks your code in the future or are you fine as-is ?

jpic:
I think so, it should document it https://github.com/yourlabs/django-autocomplete-light/blob/master/docs/api.rst#views-1

PR #1207, opened 2021-01-07
Exclude nested_admin template forms when initializing elements

### LINKED ISSUE #1174: Duplicate Initialization when using with Django Nested Admin
When I using with django-nested-admin
The nested field is being initialized twice as shown in the picture below

Version: 

```
django-nested-admin==3.3.0
django-autocomplete-light==3.7.0.dev0
```
model.py
```python
class CategoryInfo(models.Model):
    id = models.AutoField(primary_key=True)

class CategoryRelatedLabel(models.Model):
    id = models.AutoField(primary_key=True)
    label = models.ForeignKey(
        Label,
        on_delete=models.CASCADE,
        related_name="category_related_label",
        verbose_name=_("Label")
    )
    category_info = models.ForeignKey(
        CategoryInfo,
        on_delete=models.CASCADE,
        related_name = 'category_info'
    )

class CategoryRelatedLabebSubCategory(models.Model):
    id = models.AutoField(primary_key=True)
    category_related_label = models.ForeignKey(
        CategoryRelatedLabel,
        on_delete=models.CASCADE,
        related_name="category_related_label"
    )
    category = models.ForeignKey(
        Category,
        on_delete=models.CASCADE,
        related_name = 'category_related_sub_category',
        verbose_name=_("Category")
    )


```

admin.py
```python
class CategoryRelatedLabebSubCategoryInline(nested_admin.NestedTabularInline):
    model = CategoryRelatedLabebSubCategory
    form = SubCategoryForm

    extra = 0

class CategoryRelatedLabelInline(nested_admin.NestedTabularInline):
    model = CategoryRelatedLabel
    form = LabelForm
    inlines = [CategoryRelatedLabebSubCategoryInline]
    extra = 1

class CategoryInfoAdmin(nested_admin.NestedModelAdmin):
    model=CategoryInfo

    form=CategoryForm

    inlines = (CategoryRelatedLabelInline, )
```
_____________________________
Annotation 2020-08-01 120306

micjabbour:
Hello, thanks for creating this great library 😃 

Fix #1174.

This PR fixes double-initialization problems in `django-nested-admin` inline forms. It works by excluding additional id patterns before triggering element initialization. These patterns are used by the nested admin package for inserted template formsets.

I considered fixing this issue on `django-nested-admin`'s side instead, since they already [patch many client-side functions from django-admin and grappelli](https://github.com/theatlantic/django-nested-admin/blob/v3.3.3/nested_admin/static/nested_admin/src/nested-admin/utils.js#L241-L283), but I think that this would be much more complicated because in our case the patch script either has to be inserted after `autocomplete_light/autocomplete_light.js` (where `registerFunction` is defined) and before `autocomplete_light/select2.js` (where the function is used), or involve much more unnecessary hacks.


jlachowski:
This would also solve the problem of unnecessary initialisation of dynamically included formset form templates (e.g. using __prefix__ in empty form)
Thanks for this PR :)

PR #1220, opened 2021-02-19
Attempt to fix #1219

### LINKED ISSUE #1219: Using data-html has no effect when rendering new option
1. Set `data-html=true` for your select2 widget
2. Return HTML in AJAX response

**Problem**: HTML is not interpreted but displayed as text:

This seem to be related to https://github.com/yourlabs/django-autocomplete-light/issues/1171. Looks like its fix was not quite correct.
As I understand, the problem happens because of `$('<span>').text(...)` call here - https://github.com/yourlabs/django-autocomplete-light/blob/3.8.1/src/dal_select2/static/autocomplete_light/select2.js#L28 
If `data-html` is set we want to do `$(<span>).html`, not `$(<span>).text()`

eprikazc:
Attempt to fix #1219

### DIFF
--- a/src/dal_select2/static/autocomplete_light/select2.js
+++ b/src/dal_select2/static/autocomplete_light/select2.js
@@ -20,14 +20,17 @@ document.addEventListener('dal-init-function', function () {
         }
 
         function result_template(item) {
-            var text = template(item.text,
-                ($element.attr('data-html') !== undefined || $element.attr('data-result-html') !== undefined) && !item.create_id
-            );
+            var is_data_html = ($element.attr('data-html') !== undefined || $element.attr('data-result-html') !== undefined)
 
             if (item.create_id) {
-                return $('<span>').text(text).addClass('dal-create');
+              var $result = $('<span>').addClass('dal-create');
+              if (is_data_html){
+                return $result.html(item.text);
+              } else {
+                return $result.text(item.text);
+              }
             } else {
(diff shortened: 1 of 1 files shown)


jpic:
Can't we just replace the .text by a .html call ? Can you check if that works for both case ?

eprikazc:
@jpic - if we just replace `text()` with `.html()`, then text in new option will **always** be interpreted as HTML, even when `data-html` attribute is not set. I assume this is not what we want?

jpic:
I see, that might not be very safe after all, because Django allows all
unsafe characters by default in CharFields, so an injection is easy if one
just does something like `'<b>' + obj.name + '</b>`, the injection here is
obvious. This is because Django refuses to apply the OWASP security policy
that is: secure the input, not the output!

It's better to leave it this way and make this insecurity an opt-in, do you
think we also need tests or documentation?

eprikazc:
This PR does not add any new behavior, it is rather a bugfix, so no new docs are needed.. I think existing [documentation](https://django-autocomplete-light.readthedocs.io/en/master/tutorial.html#displaying-results-using-custom-html) already covers rendering of custom html quite well.
RE tests - yeah, probably adding a test or two might be helpful to prevent regressions in future. I can try to prepare them. Though, I haven't found in documentation any instructions/links on how we're supposed to run/write tests. Any hints?

PR #1233, opened 2021-04-08
replace lookbehind regex with two regex

vtbassmatt:
Safari is not so smart

fixes https://github.com/yourlabs/django-autocomplete-light/issues/1226


cnk:
Just came across this issue. Would love to see a release with this fix included.

PR #1246, opened 2021-05-13
Add event to notify of the initialization completed

andreccorrea:
Creates and dispatches the event "dal-element-initialized" to notify of the element initialization completed.

Usage:

```
$(document).on("dal-element-initialized", function(e) {
    $(`#${e.detail.element.id}`).select2('open').trigger('focus')
});
```


jpic:
Hi,

Is it possible to mention this in the documentation?

Thanks

andreccorrea:
Updated JS docs tutorial with initialization event (dal-element-initialized) example.

Issue #1260, opened 2021-09-23
[Django 4.0] the function lookup_needs_distinct  is renamed to lookup_spawns_duplicates()


pvergain:
Django 4.0
=========

- https://github.com/django/django/blob/main/docs/releases/4.0.txt

* The undocumented ``django.contrib.admin.utils.lookup_needs_distinct()``
  function is renamed to ``lookup_spawns_duplicates()``.

This function is used in https://github.com/yourlabs/django-autocomplete-light/blob/master/src/dal/views.py

jpic:
Fix in master, thanks to @fraimondo

jpic:
I believe this is fixed in 3.9.0rc2 release, please reopen if necessary.

PR #1262, opened 2021-10-05
Add support for Django 4.0

fraimondo:
Fix missing functions in Django 4.0

Fix for #1260

### DIFF
--- a/src/dal/views.py
+++ b/src/dal/views.py
@@ -6,7 +6,11 @@ from functools import reduce
 
 import django
 from django import http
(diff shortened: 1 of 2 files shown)


fraimondo:
Done!

pvergain (changes requested):
Hi, the right condition is django.VERSION >= (4, 0) instead of django.VERSION >= (3, 2).

The lookup_spawns_duplicates function does exist only for django.VERSION >= (4, 0).

Issue #1113, opened 2019-08-07
Select2 token separator can not be customized


filippo-20tab:
The `data-tags` attribute seems not to be mapped correctly in the select2 widget initialization as you can see [here](https://github.com/yourlabs/django-autocomplete-light/blob/e15864ef7eeb6d295705219cd305120bafb9e642/src/dal_select2/static/autocomplete_light/select2.js#L73). The ternary condition uses `data-tags` as condition but does not take it into account to set separators. Should it be turned into a split, in order to transform the data attribute string into an array as expected by select2?

filippo-20tab:
Actually, I was probably misled by how `data-tags` is used [here](https://github.com/yourlabs/django-autocomplete-light/blob/e15864ef7eeb6d295705219cd305120bafb9e642/src/dal_select2_taggit/widgets.py#L15). Was it meant to be just a boolean? Then, customizing the separators could be turned into a future enhancement.

jpic:
Agreed

PR #1266, opened 2021-11-08
Data token separators

fraimondo:
Fix for #1113

I need that to allow for a single tag with commas.


jpic:
Hi @fraimondo 

Thank you for your contribution! Looking forward to release this ;)

Ok to add data-token-separators, but why *require* data-tags to allow data-token-separator?

If there's no reason for that then it would be less confusing if you could remove this limitation.

Also, please fetch origin/master and rebase, fixes for django 4 have been merged already so you would not need this commit anymore (normally :joy:)

Thank you, and keep up the great work !

:tophat:

fraimondo:
I just modified the logic. The idea is to be backwards-compatible so if someone uses `data-tags`, then this will be the same as using `data-tags-separator` with a value of `[',']`

fraimondo:
Excellent! Thank you.

PR #1270, opened 2021-11-24
Fixes an issue when dal is loaded twice

FiooCode:
When dal is initialized multiple times, yl.functions stays set to {} and the select2 function is no longer present. This causes any future dal fields not to render and instead appear as normal HTML select fields. 

This error came up when I tried using HTMX in conjunction with dal. HTMX reloads the form on the fly using ajax calls. When it pulled the new form in and tried to render it, the script tags for dal were loaded again. Thus producing this error.

### DIFF
--- a/src/dal/static/autocomplete_light/autocomplete_light.js
+++ b/src/dal/static/autocomplete_light/autocomplete_light.js
@@ -3,7 +3,7 @@
  */
 
 var yl = yl || {};
-yl.functions = {};
+yl.functions = yl.functions || {};
 /**
  * Register your own JS function for DAL.
  *


jpic:
Can I ask why did you choose HTMX over unpoly.js ? Just curious, I've tried both and ended with unpoly because it had more features, such as updating two separate fragments from a request...

Anyway, nice catch! Did you know we've been patching this stuff over and over again for the last years :joy:

Hopefully DAL will be proposing a [web component based alternative soon](https://github.com/yourlabs/django-autocomplete-light/blob/alight/docs/tutorial_4.rst)

FiooCode:
I've never tried unpoly before. I'll definitely to check it out. Thanks! 

This is a great project by the way. Good work by you are everyone else who contributed. I love using this tool! Thanks for all your efforts @jpic  😄

jpic:
Thank you for the kind words! Always appreciated ;)

Did you see that [new documentation entry](https://github.com/yourlabs/django-autocomplete-light/blob/master/docs/tutorial.rst#with-djhacker) ? I hope this saves you from writing a lot of code :P

akalikin1:
Hi, seams for me the same error when I use DAL with HTMX with def clean validation,
I got duplicated DAL select field (check GIF screen capture that shows it)
When I use DAL with HTML with validation also, everything works fine, no duplicated fields

P.S. i use HTMX as convenient way to use dynamic inline formsets

dal_dup_error

FiooCode:
@akalikin1 
I haven't encountered this exact issue myself. But I have run into a similar issue where it was duplicating the select field. 

Fair warning, it's a bit of a hacky solution. I ended up with this bit of code when instantiating my form: **form(auto_id=f"field_name_{random.randint(0, 999999)}")**

if using a class-based view: 
**def get_form_kwargs(self):
        kwargs = super().get_form_kwargs()
        kwargs["auto_id"] = f"field_name_{random.randint(0, 9999999999)}_%s"
        return kwargs**

The problem is that for some reason when the same ID is used on a field and you try to instantiate the field twice, autocomplete sort of freaks out. I'm not sure if this is the same issue your running into, but hopefully it will point you in the direction of a solution. 

If you want to test it, just make sure the ID for the html field when it's rendered is different each time. After that, the field shouldn't duplicate. 

Also, please let me know if you end up finding something better. 😄 Happy coding!

akalikin1:
@FiooCode thanks for the answer

My exact problem is that when I use HTMX and render the field the class="select2-hidden-accessible" is not added to select field,
But in HTML everything is ok this class added to select field and the duplicated field is hidden

I don't know the reason and don't know how to fix it.
@jpic could you look at this problem please?

PR #1274, opened 2021-12-13
Add documentation on passing URL parameters

robertispas:
Hey,

Something I discovered very recently while working with this package was that it is indeed possible to pass parameters to the Autocomplete View through URL parameters in the widget of a form field. I've created a category on this in the tutorial as some people might find this use case helpful. It was very useful for myself and my company!

Thank you for an awesome tool,
Robert


docs/tutorial.rst
```diff
@@ -207,6 +207,51 @@ Passing options to select2
 
 .. note:: Setting a placeholder will result in generation of an an empty
           ``option`` tag, which select2 requires.
+          
+Using autocomplete with URL parameters
+======================================
+
+DAL allows the use of the Django `redirect
+<https://docs.djangoproject.com/en/4.0/topics/http/shortcuts/#redirect>`_
```
jpic, line 215: The example does not use that redirect function, did you mean reverse() ?

robertispas: Yes. Sorry, I mistyped that haha!

PR #1268, opened 2021-11-23
Django 4.0 test

jpic:
Fix Django 4 support

Do not deploy in 3.9 as this will probably break some Django 2.x support and even more whatever still works on Django 1.x

PR #1279, opened 2022-01-15
Django 4.0 imports with fallbacks

blag:
This PR fixes more imports for Django 4.0+ while still retaining the old import paths to maximize reverse compatibility.

### DIFF
--- a/src/dal_queryset_sequence/fields.py
+++ b/src/dal_queryset_sequence/fields.py
@@ -6,7 +6,10 @@ from dal_contenttypes.fields import (
 )
 
 from django import forms
-from django.conf.urls import url
+try:
+    from django.urls import re_path as url
+except ImportError:
+    from django.conf.urls import url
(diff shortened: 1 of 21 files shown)


whob13:
Is there a timeline for when this will be merged? I'd like to transition a project to 4.0 but will wait on this if it's coming soon.

Mogost:
It's better to use version checking instead of feature checking. https://adamj.eu/tech/2020/01/07/feature-checking-versus-version-checking/

blag:
> It's better to use version checking instead of feature checking. https://adamj.eu/tech/2020/01/07/feature-checking-versus-version-checking/

I would be happy to merge a PR into my branch implementing that.

whob13:
I think @Mogost makes a valid point in his comment, but perhaps this can be merged to remain compatible with Django 4 to unblock it? Thank you for the response, @blag !

blag:
Superseded by #1268. :tada:

PR #1290, opened 2022-03-09
Django version is broken

bernd-wechner:
The test project currently does not work under Django 4. Yet `requirements.txt` produces a venv that won't work because it installs Django 4. We must explicitly select 3.2 until the test project is updated to Django 4 compatibility.


jpic:
Tests now pass on Python 3.10 and Django 4 #1307, please reopen otherwise

PR #1293, opened 2022-04-22
Add the option to validate the field when creating a new choice

etiennepouliot:
Add the option of validating the data when creating a new option.

Result in the example if we try to create a new option with the added validate_slug 
```
@python_2_unicode_compatible
class TModel(models.Model):
    name = models.CharField(
        max_length=200,
        validators = [validate_slug]
    )
```
image

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -417,7 +417,7 @@ the autocomplete user interface, ie:
     urlpatterns = [
         url(
             r'^country-autocomplete/$',
-            CountryAutocomplete.as_view(create_field='name'),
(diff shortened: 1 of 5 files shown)


jpic:
Superb! But are you sure we don't want to have validate_create enabled by default?

etiennepouliot:
I didn't feel like I was the one to make that decision as this isn't my projet ;)

If you want it enabled by default I could change my PR to have an option to disable it or no option at all and enabled all the time.

what to you think ?

jpic:
Released in 3.9.5-rc1 !

Thank you <3

Will enable by default in next major release

Issue #971, opened 2018-02-14
No module named 'queryset_sequence'


Guilouf:
`django-autocomplete-light-3.2.10`
`django2.0.1`

```
INSTALLED_APPS = [
    ....
    'dal',
    'dal_select2',
    'dal_queryset_sequence',
]
```

When im trying this [tutorial](http://django-autocomplete-light.readthedocs.io/en/master/gfk.html), i can't import anything with `from dal_queryset_sequence`, not only from my own code, but also not working is source code.

EX: (in views):

`from dal_select2_queryset_sequence.views import Select2QuerySetSequenceView`

  ```
File "/home/guigui/Programmation/ZIF/gabarit/views.py", line 21, in <module>
    from dal_select2_queryset_sequence.views import Select2QuerySetSequenceView
  File "/usr/local/lib/python3.6/dist-packages/dal_select2_queryset_sequence/views.py", line 3, in <module>
    from dal_queryset_sequence.views import BaseQuerySetSequenceView
  File "/usr/local/lib/python3.6/dist-packages/dal_queryset_sequence/views.py", line 7, in <module>
    from queryset_sequence import QuerySetSequence
ModuleNotFoundError: No module named 'queryset_sequence'
```

I installed the package via pip, its not clear in the docs, do i need to install a a new django extension, is it [this](https://pypi.python.org/pypi/django-querysetsequence) queryset_sequence ??

Guilouf:
Ok, `queryset_sequence` was, already or as a autocomplete dependency, installed, but had to be upgraded, via `python3.6 -m pip install django-querysetsequence --upgrade`

jpic:
Thanks for reporting!

NaturalBornCamper:
Might I suggest the doc be updated to say we have to install:
`django-querysetsequence==0.13`?
I didn't know that package and I thought it was included in dal when the docs say it should be added in `INSTALLED_APPS`, I struggled quite a bit before finding it in an example's requirements.txt

tylerecouture:
Ran into this too. Docs [here](https://django-autocomplete-light.readthedocs.io/en/master/gfk.html#register-the-view-for-the-form) don't mention that this is a separate package that has to be installed.

PR #1297, opened 2022-06-12
Add reminder to install django-querysetsequence

tylerecouture:
This mirrors the reminder elsewhere in the documentation "# don't forget to pip install djhacker".  Several issues have been started in the past due to the error that's caused by forgetting (or not understanding) that this isn't installed by default.  e.g. #971.  Just happened to me now.

### DIFF
--- a/docs/gfk.rst
+++ b/docs/gfk.rst
@@ -61,7 +61,7 @@ Result:
 
 .. code-block:: python
 
-    from dal import autocomplete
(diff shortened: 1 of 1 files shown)


jpic:
Thanks!

jpic:
Also, note that if you `pip install django-autocomplete-light[gfk]` then it will pull django-querysetsequence

PR #1306, opened 2022-09-06
Test Django 4.1

adamchainz:
This PR updates the tox and travis config files to drop old versions of Django and Python, and test Django 4.1.

I tried running the tests locally. The unit tests pass, but those using splinter fail. It seems pytest-splinter is incompatible with some changes to splinter. I tried rolling back splinter as far as version 0.14, but that only got me different errors. The errors with the latest splinter are all:

```
    @pytest.fixture(scope="session")
    def browser_patches():
        """Browser monkey patches."""
        patch_webdriver()
>       patch_webdriverelement()

../.tox/py310-dj41/lib/python3.10/site-packages/pytest_splinter/plugin.py:278:
_ _ _ 
    def patch_webdriverelement():  # pragma: no cover
        """Patch the WebDriverElement to allow firefox to use mouse_over."""

        def mouse_over(self):
            """Perform a mouse over the element which works."""
            (
                ActionChains(self.parent.driver)
                .move_to_element_with_offset(self._element, 2, 2)
                .perform()
            )

        # Apply the monkey patch for Firefox WebDriverElement
>       firefox.WebDriverElement.mouse_over = mouse_over
E       AttributeError: module 'splinter.driver.webdriver.firefox' has no attribute 'WebDriverElement'
```

I don't think Travis is still running - it hasn't reported any status on this PR. It may need just re-setting up, to use the GitHub app rather than the old token integration. It would be good to move to GitHub Actions anyway, since that's better maintained.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,18 @@ matrix:
     env: TOXENV=checkqa
   - python: 3.7
     env: TOXENV=docs
+  - python: 3.6
+    env: TOXENV=py36-dj32
   - python: 3.7
-    env: TOXENV=py37-dj30
-  allow_failures:
-  - env: TOXENV=py37-dj30
+    env: TOXENV=py37-dj32
+  - python: 3.8
+    env: TOXENV=py38-dj40
+  - python: 3.8
+    env: TOXENV=py38-dj41
+  - python: 3.9
+    env: TOXENV=py39-dj41
+  - python: 3.10
+    env: TOXENV=py310-dj41
 install:
 - travis_retry pip install -U pip
 - travis_retry pip install tox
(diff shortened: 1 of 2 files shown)


jpic:
Apparently so yes https://github.com/pytest-dev/pytest-splinter/issues/169#issuecomment-1242379232

jpic:
@mpasternak yes, you fixed it in 0f9cae60
But it's not in pypi

jpic:
I believe py310 and dj4 are supported by https://pypi.org/project/django-autocomplete-light/3.9.5rc0/

Can someone please confirm this?

Thanks!

adamchainz:
Thank you @jpic . When I'm back working with my client who uses django-autocomplete-light, in a week, I'll test this (unless I get distracted by Djangocon Europe).

PR #1335, opened 2023-08-16
fix: Fix event format for Django 4.2

last-partizan:
Event format has been changed for Django 4.2, and old variant isn't working.

https://github.com/django/django/commit/eabc22f919e6c1774842e628400b87ac56c47bce

This PR fixes it, but we need to decide how to handle the old django versions.

Cleaner variant would be to support only latest django version, and those with old version should use old library versions.

If we want to support all django versions, i need to modify this, to check if `$newFormset` is defined or not, and then use two different call variants.

What do you think?


jpic:
Should be fine, thank you!

jpic:
Please try 3.10.0-rc3 and let me know if it works for you

last-partizan:
Yes, it's fixed. No more errors when adding inline.

Issue #1133, opened 2020-01-09
Created tags are named `Create "tag_text"`


PeteAndersen:
Using DAL + django-taggit. When tag creation is enabled, the created tags are including the `Create "<text>"` boilerplate from the widget in the tag name. For example, if I type `foo`, and a tag named foo does not exist, the widget gives me the option `Create "foo"`. Clicking on it sends a POST request to the autocomplete view with the text field set literally to `Create "foo"` instead of just `foo`, and the tag is created with the full text.

jpic:
I think you don't need the Create function with taggit, because taggit will always create new tags.

PR #1336, opened 2023-09-08
Update taggit.rst docs.

voberoi:
Hi! I ran into this issue and figured out a fix. Maybe this is not the correct change to make or document -- if it isn't, please let me know how to address this in my app + I'd be happy to update the PR to address the underlying code instead if that's the issue.

---

For `taggit` the create option is broken, and it is unnecessary.

One of the drop down options includes the new tag. So if you've typed `funny` and `funny` does not exist you'll have:

```
funny
Create 'funny'
```

Picking `funny` works as intended -- it creates a new tag properly.

Picking `Create 'funny'` creates two tags and labels them wonkily.

This change removes the create option entirely from the docs so this works out of the box for end-users.

### DIFF
--- a/docs/taggit.rst
+++ b/docs/taggit.rst
@@ -48,6 +48,9 @@ Select2 and a QuerySet of Tag objects:
 
             return qs
 
+        def get_create_option(self, context, q):
+            return []
+
 Don't forget to :ref:`register-view`.
 
 .. note:: For more complex filtering, refer to official documentation for


voberoi:
Seems like there was an issue in 2020 but it manifested itself differently: https://github.com/yourlabs/django-autocomplete-light/issues/1133

In my case, the creation event creates two tags, and the one that ends up being displayed contains a tag ID in the name. I haven't made any changes to my implementation and have done what's in the docs verbatim.

There's also comment in that issue that states the create option probably isn't necessary for taggit.

jpic:
Great! Thanks a heap!

jpic:
Please try 3.10.0-rc3 and let me know if it works for you

PR #1324, opened 2023-04-25
Test on Django 4.2

adamchainz:

PR #1344, opened 2023-11-06
Test on Django 5.0

adamchainz:
Add official support for this [new Django version](https://docs.djangoproject.com/en/5.0/releases/5.0/), currently in beta. Based on the changes needed back in #1324.

Still working on getting tests running cleanly locally, looks like there are new warnings to fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,12 @@ matrix:
     env: TOXENV=py39-dj42
   - python: 3.10
     env: TOXENV=py310-dj42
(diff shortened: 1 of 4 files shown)


adamchainz:
After getting things running, I found a bunch of test failures. I don't feel like I have enough context to fix them all unfortunately!

jpic:
Thanks! Could fix them except for django-nested-admin, so good old jquery conflict for anyone interested to contribute.

PR #1368, opened 2024-07-12
Fix - Object of type ValidationError is not JSON serializable

Spawin:
On [Creation of new choices in the autocomplete form](https://django-autocomplete-light.readthedocs.io/en/master/tutorial.html#creation-of-new-choices-in-the-autocomplete-form), when we use `validate_create=True` and the validation failed we have `TypeError: Object of type ValidationError is not JSON serializable`.

Corrections made :
- fix the TypeError
- return of the error of the concerned field to the view

Versions used:
- django==5.0.7
- django-autocomplete-light==3.11.0


Spawin:
Please i need review here

PR #1383, opened 2025-01-30
build and release wheel

dimbleby:
Direct invocation of setuptools is [deprecated](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/).  Build and publish a wheel.

I did not see where you set up dependencies for build/publish: you will need to `pip install build` before building (just as you need to `pip install twine` before publishing).

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -36,7 +36,7 @@ echo -e "$(python changelog.py $1)\n$(cat CHANGELOG)" > CHANGELOG
 git add setup.py docs/conf.py CHANGELOG
 git commit -m "Release $1"
 git tag $1
-python setup.py sdist
+python -m build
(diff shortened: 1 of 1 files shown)


jpic:
Seems legit (by PyPA), I wonder how that works with setupmeta ... will soon see on my other repos
Thanks for the bot, nice gesture :angel:

PurpleToti:
The changes from this PR were integrated into yourlabs/django-autocomplete-light via PR #1407 (Modernise for Python 3.11-3.14 / Django 5.2-6.0+), merged 2026-04-21. Thank you for the contribution! The build was also extended to publish both the sdist and the wheel (the original PR only uploaded the tarball).

